PR #3, opened 2025-04-01
docs: add SDK design doc template

rwalworth:
**Description**:
This PR adds the template for the design documents that should be added to this repo.

See [here](https://github.com/hiero-ledger/hiero/pull/48) for other discussion points on this.

### DIFF
--- a/templates/sdk-design-doc-template.md
+++ b/templates/sdk-design-doc-template.md
@@ -1,0 +1,86 @@
+# Hiero SDK Design Doc Template
+
+This is the template to use for Hiero SDK design documents. These documents should be generated for every new feature looking to be added to the Hiero SDKs. These features can include, but are not limited to, new HIP implementations, internal SDK processing changes, other new or changed APIs, etc. Any change that should be applied to all Hiero SDKs should use this design doc template to describe the proposed changes and test plan.
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


templates/sdk-design-doc-template.md
```diff
@@ -0,0 +1,86 @@
...
+
+Create a numbered list of end-to-end tests that should be implemented to fully test the new feature design. These should use the Given/When/Then idiom in order to fully define the test prerequisites and setup, what’s being tested, and the expected response. The tests should invoke all new response codes and validate them.
+
+1. Given `<prerequisites>`, when `<action>`, then `<result>`.
+2. Given `<prerequisite #1>`, `<prerequisite #2>`, and `<prerequisite #3>`, when `<action>`, then `<result #1>`, `<result #2>`, and `<result #3>`.
+3. …
+
+### TCK
+
+The tests defined above should also be defined in the [TCK repository](https://github.com/hiero-ledger/hiero-sdk-tck). Issues should be created in the TCK repository based on this content to document all end-to-end tests and link it as reference in the design doc.
```
0xivanov, line 80: So from now on, as we add new integration tests, we will add the same flows to the TCK?
Overall, the TCK is quite slower and does not have all of the internal utils we use it our e2e tests. There might be some edge cases where the tests on the TCK would be tricky to pull off.

0xivanov (approved):
Looks good overall!

templates/sdk-design-doc-template.md
rwalworth: I think it makes sense that there's a lot of crossover between SDK e2e tests and the TCK. It may not be that every test could be brought over to the TCK, but would think most of them should.

PR #18, opened 2025-05-08
HIP-1046 design

0xivanov:
**Description**:
Adds HIP-1046 design for SDKs

### DIFF
--- a/hips/hip-1046.md
+++ b/hips/hip-1046.md
@@ -20,55 +20,60 @@ This HIP introduces new grpc web proxy field in the address book schema, making 
   - `Endpoint getGrpcWebProxyEndpoint()`
   - `NodeUpdateTransaction setGrpcWebProxyEndpoint(Endpoint)`
 
-### AddressBookQuery
-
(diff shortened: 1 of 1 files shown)


hips/hip-1046.md
```diff
@@ -0,0 +1,119 @@
...
+	PublicKey   string
+	AccountID   AccountID
+	NodeID      int64
+	CertHash    []byte
+	Addresses   []Endpoint
+	Description string
+}
+```
+
+Each node has a list of `Endpoint` where I’m guessing the grpc data will be returned. We only need to determine if it’s grpc endpoint or web proxy endpoint. This could be achieved with enum defined in the protobufs, but we don’t have the mirror node spec yet.
```
rwalworth, line 40: Do we have the mirror node spec? Can we update this?

venilinvasilev: I've made the adjustments, it should be more clear now :).

ivaylonikolov7 (approved):

PR #23, opened 2025-05-21
docs: client static network initialization should use `AddressBookQuery`

rwalworth:
**Description**:
This PR proposes a different way of initializing static Client networks. Instead of using hardcoded values or files to read from, the Client should just use an `AddressBookQuery` under the hood to always grab the most up-to-date network values. This will help prevent issues of updating SDK versions and provide a more stable developer experience.

### DIFF
--- a/proposals/client-for-network-uses-mirror-node-address-book.md
+++ b/proposals/client-for-network-uses-mirror-node-address-book.md
@@ -1,0 +1,54 @@
+# Client Initialization for Static Networks Uses Address Book Mirror Node Query
+
+**Date Submitted:** 2025-05-21
+
+## Summary
+
+This proposal describes a change in client network initialization for static networks (e.g. Hedera mainnet). Currently, the Hiero SDKs read a static file (or hardcoded values) that describe the IPs, ports, and account IDs of the nodes running the network. The SDKs will then wait roughly ten seconds, then query the network mirror node to grab the most up-to-date address book. This can be problematic for a couple of reasons:
+
+- Network updates require an updated file from which the SDKs read.
+- The first ten seconds of operation could use an outdated network, resulting in execution errors.
+
+This proposal suggests removing this method of static network initialization and replacing it with a mirror node query. This will prevent needing to manually update the SDKs with new networks, as well as prevent invalid node errors if attempting execution within the first ten seconds of operation.
+
+## Internal Changes
+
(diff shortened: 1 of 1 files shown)


proposals/client-for-network-uses-mirror-node-address-book.md
```diff
@@ -0,0 +1,54 @@
...
+                internal::Network::getNetworkFromAddressBook(
+                    AddressBookQuery().execute(client),
+                    internal::BaseNodeAddress::PORT_NODE_PLAIN)));
+    return client;    
+}
+```
+
+The implementation will still require some sort way to read the mirror node network. However, the SDKs already do this so nothing needs to change there. The SDKs should use the established mirror node network to retrieve the address book for the consensus node network, and use the address book to initialize its consensus node network.
+
+With this update, the ten second scheduled update for Hiero SDK networks can be removed. Since the SDK networks will be initialized from the start with the most up-to-date address book information, there is no longer a good reason to do another one-time update ten seconds later.
```
SimiHunjan, line 34: what would happen if the SDK could not retrieve the address book? Do we need think about something where if in x time we don't receive the address book from the mirror node the SDK does y? Part of the answer would be we can get the address book from multiple sources, like multiple running mirror nodes, but we don't exactly have that today.

proposals/client-for-network-uses-mirror-node-address-book.md
```diff
@@ -0,0 +1,54 @@
...
+    client.mImpl->setMirrorNetwork(internal::MirrorNetwork::forMainnet());
+    client.mImpl->mNetwork =
+        std::make_shared<internal::Network>(
+            internal::Network::forNetwork(
+                internal::Network::getNetworkFromAddressBook(
+                    AddressBookQuery().execute(client),
+                    internal::BaseNodeAddress::PORT_NODE_PLAIN)));
+    return client;    
+}
+```
```
venilinvasilev, line 30: I think this is the most logical approach to handle this scenario. We have thinked about this with our team. The problem we see with this is that in the JS SDK we need to make the instantiation of the Client asynchronous (because the query operation is asynchronous) which is a breaking change and should be done in a major version of it.

rwalworth: Hmm maybe other SDKs that don't require a breaking change can implement but the ones that break (such as JS) can wait until a v3?

venilinvasilev: I'm fine with this approach.

PR #57, opened 2025-10-08
Api guideline

hendrikebbers:
This pull request introduces a new API guideline document to the repository. The document establishes conventions and best practices for designing and documenting APIs for SDKs, with a strong emphasis on language-agnostic documentation syntax. This aims to improve consistency, usability, and maintainability of APIs across projects.


guides/api-guideline.md
```diff
@@ -0,0 +1,135 @@
...
+The following annotations should be used:
+- `@immutable`: Indicates that the field is immutable and cannot be changed after creation.
+- `@optional`: Indicates that the field is optional and can be null or undefined.
+- `@default(value)`: Indicates that the field has a default value.
+- `@deprecated`: Indicates that the field is deprecated and should not be used.
+- `@removed`: Indicates that the field must be removed.
+- `@min(value)`: Indicates the minimum value for numeric fields.
+- `@max(value)`: Indicates the maximum value for numeric fields.
+- `@minLength(value)`: Indicates the minimum length for string fields.
+- `@maxLength(value)`: Indicates the maximum length for string fields.
```
rwalworth, line 57: Does the behavior need to specified if a value is provided lower than the min/above the max? Or should it not even be allowed?

guides/api-guideline.md
```diff
@@ -0,0 +1,135 @@
...
+- `@pattern(regex)`: Indicates a regex pattern that the string field must match.
+
+### Method annotations
+
+Method annotations can be used to provide additional information about methods.
+The following annotations should be used:
+- `@async`: Indicates that the method is asynchronous and returns a promise or future.
+- `@throws(error-type)`: Indicates that the method can throw an exception/error.
+- `@deprecated`: Indicates that the method is deprecated and should not be used.
+- `@removed`: Indicates that the method must be removed.
```
rwalworth, line 67: I'm trying to think of a time when `@deprecated` or `@removed` would ever be used in our design docs (same for attributes above too). If we are deprecating/removing methods, I think it would make sense to explicitly call them out under a `Deprecate/Remove` heading or something

guides/api-guideline.md
rwalworth: Should this be enforced at the SDK level or consensus node level? For example, memos are currently limited to 100 bytes. Enforcing at SDK level would prevent users from ever inputting >100 byte memos and make it fail faster. However, what if consensus node decides to up this to 150 bytes? Users would need to update SDKs to get this new enforcement. Whereas if we just allowed enforcement at the consensus node level always, SDKs wouldn't need to bother (and descriptive `MEMO_TOO_LONG` response codes would still highlight the exact issue).

guides/api-guideline.md
```diff
@@ -0,0 +1,131 @@
...
+
+The following basic data types should be used in the API documentation.
+
+| Data Type       | Description                                                   |
+|-----------------|---------------------------------------------------------------|
+| `string`        | A sequence of characters                                      |
+| `intX`          | A signed integer of X bits                                    |
+| `uintX`         | An unsigned integer of X bits                                 |
+| `int`           | A signed integer of 32 bits                                   |
+| `long`          | A signed integer of 64 bits                                   |
```
rwalworth: It seems a bit unnecessary to me to have `int` and `long` when we have `intX` and `uintX`, which are just `int32` and `int64`

guides/api-guideline.md
```diff
@@ -0,0 +1,131 @@
...
+
+Complex data types can be defined using the basic data types and other complex data types.
+The following syntax should be used to define complex data types:
+```
+DataTypeName {
+    fieldName1: DataType1
+    fieldName2: DataType2
+    ReturnType methodName(param1: DataType1, param2: DataType2)
+}
+```
```
rwalworth: Should `public` and `private` be specified? Or should everything defined be assumed to be `public` since anything but is just implementation details?

guides/api-guideline.md
```diff
@@ -0,0 +1,131 @@
...
+User {
+    @immutable id: uuid
+    @minLength(3) @maxLength(100) email: string
+    @optional @min(0) @max(120) age: uint8
+    
+    @async
+    @throws(not-found-error)
+    @optional UserProfile fetchProfile(apiKey: string)
+}
+```
```
rwalworth: Should we provide specs a type extending another? Something like
```
Derived extends Base {
  ...
}
```

guides/api-guideline.md
hendrikebbers: Everything that is defined must be public. We should not define global private methods

guides/api-guideline.md
```diff
@@ -0,0 +1,131 @@
...
+
+Enumerations can be defined using the following syntax:
+```
+enum EnumName {
+    VALUE1
+    VALUE2
+}
+```
+
+### Attribute annotations
```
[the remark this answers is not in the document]
hendrikebbers: I will think about that and how we can define it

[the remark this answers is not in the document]
hendrikebbers: Can we add that in a future step?

guides/api-guideline.md
hendrikebbers: I'm still not fully convinced on not having int and long but happy to remove it for now to get the PR merged

PR #66, opened 2025-10-31
create key always by defining algorithm

hendrikebbers:
This pull request refactors the key creation and decoding logic in the SDK and sample applications to require explicit specification of `KeyAlgorithm` and `KeyEncoding` when constructing `PrivateKey` and `PublicKey` objects. This improves type safety and clarity in cryptographic operations. The changes also update sample usage to match the new method signatures and add utility functions for key decoding.

### DIFF
--- a/v3-sandbox/prototype/sdk/src/main/java/org/hiero/sdk/simple/network/keys/PublicKey.java
+++ b/v3-sandbox/prototype/sdk/src/main/java/org/hiero/sdk/simple/network/keys/PublicKey.java
@@ -46,7 +46,7 @@ public interface PublicKey extends Key {
      */
     //TODO: Do we really want to have that method without the definition of the KeyAlgorithm and KeyEncoding?
     @NonNull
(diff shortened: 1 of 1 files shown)


v3-sandbox/prototype/sdk/src/main/java/org/hiero/sdk/simple/internal/network/key/KeyFactory.java
```diff
@@ -101,15 +32,10 @@ public static PrivateKey createPrivateKey(@NonNull final byte[] privateKey) {
      * @throws IllegalArgumentException if the bytes do not match the algorithm
      */
     @NonNull
-    public static PrivateKey createPrivateKey(@NonNull final byte[] privateKey, @NonNull final KeyAlgorithm algorithm) {
+    public static PrivateKey createPrivateKey(@NonNull final KeyAlgorithm algorithm, @NonNull final byte[] privateKey) {
```
rwalworth: So we're enforcing that the user needs to know what type of key they're inputting? Do we not want to allow the SDK to determine the key type itself based on DER encoding?

PR #80, opened 2025-11-11
Update based on API guideline changes

hendrikebbers:
This pull request updates the prototype API model definition syntax to improve consistency and clarity. The main change is the replacement of the `@immutable` and `@abstraction` annotations with a new syntax using `@@immutable` and direct keyword usage, respectively. This affects multiple data types and key abstractions.


rwalworth (approved):
LGTM

PR #91, opened 2025-11-19
SDK sample code guide

rwalworth:
**Description**:
This PR adds an SDK sample code guide that establishes a standard for how SDK examples should be written.

### DIFF
--- a/guides/code-sample-guide.md
+++ b/guides/code-sample-guide.md
@@ -1,0 +1,158 @@
+# Reference Hiero SDK Sample Code Guide
+
+This guide defines the standard structure and conventions for writing Hedera SDK code examples across SDK repositories, blogs, developer documentation, and the Hedera Portal Playground.
+
+It is intended for developers contributing to the SDK or creating sample code for docs and Playground environments.
+
+Each example must use ECDSA (w EVM Alias from Public Key) as the default key type, support both .env and hardcoded credentials, follow a consistent step structure, include debug logging and HashScan links, and close the client gracefully.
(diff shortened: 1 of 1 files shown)


guides/code-sample-guide.md
```diff
@@ -0,0 +1,158 @@
...
+Each example must use ECDSA (w EVM Alias from Public Key) as the default key type, support both .env and hardcoded credentials, follow a consistent step structure, include debug logging and HashScan links, and close the client gracefully.
+
+
+## Quick Reference Standards
+
+| **Element**                | **Standard**                                |
+|----------------------------|---------------------------------------------|
+| Default key type           | Always ECDSA w/ EVM Address from Public key |
+| Operator setup             | `.env` or hardcoded fallback                |
+| Environment variable names | `HEDERA_OPERATOR_ID`, `HEDERA_OPERATOR_KEY` |
```
venilinvasilev:
```suggestion
| Environment variable names | `OPERATOR_ID`, `OPERATOR_KEY` |
```

Reccetech: So Patches pointed out that in a multichain world often you'll have something like an env file with different keys for different blockchains. So I don't know how big this use case is - but at the same time I think it cannot hurt.

rwalworth:
> This is a JavaScript only guide as far as I can say. We should make that clear and provide similar guidelines for the other languages.

@Reccetech would you want to add examples scripts for each language in this doc, or create several different docs for each language?

Issue #34, opened 2025-06-16
Create example json configuration


rwalworth:
This would be an example `Client` JSON configuration a user could input into `Client.fromConfig`

KubanjaElijahEldred:
hlo am interested please assign me this issue

KubanjaElijahEldred:
thankyou so much for the opportunity granted, i will contact you in case of
anything

Priyanshubhartistm:
@rwalworth 
Hi, I would like to work on this issue. Please assign it to me.

rwalworth:
You are assigned @Priyanshubhartistm, good luck and please reach out if you have any questions!

Priyanshubhartistm:
Hi @rwalworth, thank you for assigning this issue. I have completed the requested documentation update and opened the linked PR. The changes are scoped only to Issue #34 (Client.fromConfig JSON configuration example with brief parameter explanations). Could you please review it when you get a chance? Thank you!

PR #99, opened 2025-11-25
Add Client JSON configuration example documentation

KubanjaElijahEldred:
- Created comprehensive JSON configuration guide for Client.fromConfig()
- Includes all supported parameters: network, mirrorNetwork, shard, realm, grpcDeadline, requestTimeout
- Provides network-specific examples for mainnet, testnet, and previewnet
- Contains usage examples for JavaScript/TypeScript, Java, and Go SDKs
- Includes best practices and parameter explanations
- Addresses issue #34: Create example JSON configuration

**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

### DIFF
--- a/guides/client-json-configuration-example.md
+++ b/guides/client-json-configuration-example.md
@@ -25,15 +25,16 @@ This document provides an example JSON configuration that can be used with `Clie
 
 ### Network Configuration
 
-- **`network`** (object, required): Maps consensus node addresses to their corresponding account IDs
+- **`network`** (object, optional): Maps consensus node addresses to their corresponding account IDs
   - **Key format**: `"IP:PORT"` (e.g., `"35.242.233.154:50211"`)
   - **Value format**: `"SHARD.REAL.ACCOUNT"` account ID (e.g., `"0.0.3"`)
   - **Purpose**: Defines the consensus nodes that the client will connect to for transaction submission
+  - **Note**: If not provided, the client can query the address book from mirror nodes to discover network nodes
 
-- **`mirrorNetwork`** (array or string, optional): Mirror node endpoints for query operations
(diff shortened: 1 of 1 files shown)


guides/client-json-configuration-example.md
```diff
@@ -0,0 +1,181 @@
...
+}
+```
+
+## Configuration Parameters
+
+### Network Configuration
+
+- **`network`** (object, required): Maps consensus node addresses to their corresponding version numbers
+  - **Key format**: `"IP:PORT"` (e.g., `"35.242.233.154:50211"`)
+  - **Value format**: `"MAJOR.MINOR.PATCH"` version string (e.g., `"3.5.10"`)
```
rwalworth: This should be the value of the account ID for that node (so `3.5.10` would be the associated account in this example)

guides/client-json-configuration-example.md
```diff
@@ -0,0 +1,181 @@
...
+  - **Purpose**: Defines the consensus nodes that the client will connect to for transaction submission
+
+- **`mirrorNetwork`** (array or string, optional): Mirror node endpoints for query operations
```
rwalworth: I think we should always have this as an array

guides/client-json-configuration-example.md
```diff
@@ -0,0 +1,181 @@
...
+  "grpcDeadline": 10000,
+  "requestTimeout": 120000
+}
+```
+
+## Configuration Parameters
+
+### Network Configuration
+
+- **`network`** (object, required): Maps consensus node addresses to their corresponding version numbers
```
rwalworth: I'm not sure if we want to keep `network` as required. It may be the case that a user only has a mirror node endpoint, but they could still use that to query for the address book and get the network configuration

PR #133, opened 2026-02-17
collections in swift

hendrikebbers:
This pull request adds a new section to the Swift API best practices guide, focusing on how to handle collections in public APIs. The section explains why collections should never be returned as `nil`, highlights Swift's value semantics for collections, and provides code examples for best practices.


guides/api-best-practices-swift.md
```diff
@@ -33,4 +33,73 @@ For types with only `@@immutable` fields, use `struct` with `let` properties:
...
-}
\ No newline at end of file
+}
+```
+
+## Collections
+
+The public API must never return `nil` for collections. Instead, an empty collection must be returned.
+Since the data types `list`, `set`, and `map` should never be nullable, it never makes sense to wrap a collection in an
+`Optional` in the public API.
```
rwalworth: > The public API must never return `nil` for collections.

"Never" is pretty strict here. In practice, `nil` and an empty collection can have different semantics - for example, in an update transaction, `nil` means "the user is not trying to update this collection" while an empty collection means "the user wants to clear everything from it." Wrapping a collection in `Optional` is the idiomatic way to express that distinction in Swift.

I'd soften this to say collections should not be `nil` *when representing the result of a query or a read* (return an empty collection instead), but acknowledge that `Optional` collections are appropriate when `nil` carries distinct meaning (e.g., "no change" vs. "clear").

PR #136, opened 2026-02-23
docs: Add HIP-1137 SDK design document

rwalworth:
## Summary

This PR adds the SDK design document for [HIP-1137](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-1137.md) (Block Node Discoverability via On-Chain Registry). The document specifies the new types, transactions, mirror node query API, updated APIs, test plan, and SDK example that each SDK team needs to implement HIP-1137 consistently across all Hiero SDKs.

**Key Changes:**
- Define an inheritance-based endpoint hierarchy: `RegisteredServiceEndpoint` (abstract base) with `BlockNodeServiceEndpoint`, `MirrorNodeServiceEndpoint`, and `RpcRelayServiceEndpoint` subtypes
- Define `BlockNodeApi` enumeration for well-known block node endpoint APIs
- Specify three new transactions: `RegisteredNodeCreateTransaction`, `RegisteredNodeUpdateTransaction`, and `RegisteredNodeDeleteTransaction`
- Define mirror node query types: `RegisteredNode`, `RegisteredNodeAddressBook`, and `RegisteredNodeAddressBookQuery` (following `AddressBookQuery`/`NodeAddressBook` pattern)
- Document updates to `TransactionReceipt`, `NodeCreateTransaction`, and `NodeUpdateTransaction`
- Document schedulability of all three new transactions
- Include an 18-case test plan and a full lifecycle SDK example (create → query → update → associate → delete)

## Motivation

HIP-1137 introduces an on-chain registered node registry to the Hiero ledger, enabling block node operators, mirror node operators, and RPC relay operators to publish their nodes so that clients can discover them. The SDKs need a cross-language design document that specifies the new types, transactions, updated APIs, and testing requirements so that each SDK team can implement the feature consistently.

---

### Endpoint Hierarchy

| Type | Description |
| --- | --- |
| BlockNodeApi | Enum of well-known block node endpoint APIs (OTHER, STATUS, PUBLISH, SUBSCRIBE\_STREAM, STATE\_PROOF) |
| RegisteredServiceEndpoint | Abstract base type declaring address (IP or FQDN), port, and TLS requirement |
| BlockNodeServiceEndpoint | Extends RegisteredServiceEndpoint with `endpointApi: BlockNodeApi` |
| MirrorNodeServiceEndpoint | Extends RegisteredServiceEndpoint (empty subtype, extensible for future fields) |
| RpcRelayServiceEndpoint | Extends RegisteredServiceEndpoint (empty subtype, extensible for future fields) |

> **Design rationale -- inheritance vs. composition:** The endpoint hierarchy uses inheritance rather than the protobuf `oneof`-based composition. `RegisteredServiceEndpoint` is an abstract base type with three concrete subtypes, one per node service kind. This lets the type system enforce that each endpoint is exactly one kind without runtime `oneOf` checks or marker-object fields. It follows the same pattern the SDKs already use for `TransactionBody`, where the protobuf `oneof` is split into separate transaction classes. `MirrorNodeServiceEndpoint` and `RpcRelayServiceEndpoint` are intentionally defined as empty subtypes so that type-specific fields can be added in future HIPs without restructuring the hierarchy.

### New Transactions

| Transaction | Description |
| --- | --- |
| RegisteredNodeCreateTransaction | Creates a registered node with an admin key, optional description, optional node account ID, and service endpoints; returns a registeredNodeId in the receipt |
| RegisteredNodeUpdateTransaction | Updates description, admin key, node account ID, or service endpoints of an existing registered node |
| RegisteredNodeDeleteTransaction | Removes a registered node from the address book |

### Mirror Node Query API

| Type | Description |
| --- | --- |
| RegisteredNode | Immutable representation of a registered node as stored in network state |
| RegisteredNodeAddressBook | Collection of `RegisteredNode` objects, following the `NodeAddressBook` pattern |
| RegisteredNodeAddressBookQuery | Queries the mirror node for registered nodes, following the `AddressBookQuery` pattern |

> **Note:** The query API defines the SDK-facing contract only. Implementation details (transport, endpoint path, pagination, filtering) depend on the mirror node implementation and will be added at a later date if needed. SDKs should not implement this query until the mirror node API is available.

### Updated APIs

| API | Change |
| --- | --- |
| TransactionReceipt | New `registeredNodeId` field (immutable, nullable uint64) |
| NodeCreateTransaction | New `associatedRegisteredNodes` list and `addAssociatedRegisteredNode` method |
| NodeUpdateTransaction | New `associatedRegisteredNodes` list and `addAssociatedRegisteredNode` method |

### Internal Changes

- **Schedulability:** All three registered node transactions are schedulable via `ScheduleCreateTransaction`. The concrete implementation is SDK-specific.
- **Response Codes:** HIP-1137 does not define any new response codes. If response codes specific to registered node transactions are introduced during consensus node implementation, this document will be updated.

### Test Plan

The design document includes 18 test cases organized across six categories:

| Category | Tests | Coverage |
| --- | --- | --- |
| Create (happy path) | 1-5 | Block node, mirror node, RPC relay, mixed endpoints, with description |
| Create (failure) | 6-7 | Missing admin key, empty endpoints |
| Update | 8-12 | Description, endpoints, admin key rotation, missing signature, non-existent ID |
| Delete | 13-15 | Successful delete, already-deleted, non-existent ID |
| Association | 16-17 | NodeCreateTransaction and NodeUpdateTransaction with associatedRegisteredNodes |
| Endpoint switch | 18 | IP address to domain name |

### SDK Example

A full lifecycle example walks through: create → query → update → associate with consensus node → delete.

---

## Testing

**Test Plan:**
- Verify the design document covers all protobuf messages defined in HIP-1137
- Verify all three new transactions (`RegisteredNodeCreate`, `RegisteredNodeUpdate`, `RegisteredNodeDelete`) are specified
- Verify the endpoint hierarchy matches the inheritance-based design (abstract base + three subtypes)
- Verify mirror node query types (`RegisteredNode`, `RegisteredNodeAddressBook`, `RegisteredNodeAddressBookQuery`) are documented
- Verify updated APIs (`TransactionReceipt`, `NodeCreateTransaction`, `NodeUpdateTransaction`) are documented
- Verify schedulability of all three transactions is noted
- Verify the test plan addresses create, update, delete, failure, and association scenarios
- Verify the SDK example covers the full registered node lifecycle including the query step

## Files Changed Summary

| Category | Files | Notes |
| --- | --- | --- |
| Proposals | proposals/hips/hip-1137.md | New SDK design document (+282 lines) |

## Breaking Changes

**None.** This PR is documentation-only. HIP-1137 itself introduces net-new functionality with no backwards compatibility impact on existing APIs.

### DIFF
--- a/proposals/hips/hip-1137.md
+++ b/proposals/hips/hip-1137.md
@@ -50,6 +50,8 @@ abstraction RegisteredServiceEndpoint {
 }
 ```
 
+> An additional `type: string` field should be added for JS to allow a user to determine the endpoint type.
+
 #### BlockNodeServiceEndpoint
 
 A registered service endpoint for a block node. Extends the base endpoint with the specific block node API that the endpoint exposes.


jsync-swirlds (approved):

guides/api-guideline.md
```diff
@@ -238,6 +238,12 @@ The following annotations should be used:
   at the SDK level.
 - `@@pattern(regex)`: Indicates a regex pattern that the string field must match. Should be included if the value must
   be enforced at the SDK level.
+- `@@getSet`: The field must not be exposed as a public property. The public API exposes a getter and a setter instead.
```
hendrikebbers: My assumption is that every field defined by using that guideline is somehow part of the public API. With that it is either @@immutable and only provides a `get` functionality or it provides a `get` and a `set` functionality.

How would a property be exposed from your point of view of `@@getSet` is not defined?

proposals/hips/hip-1137.md
```diff
@@ -0,0 +1,239 @@
...
+
+Describes an RPC relay endpoint. Currently carries no additional data.
+
+```
+@@finalType
+RpcRelayEndpoint {
+}
+```
+
+> **Design rationale -- marker types vs. enum:** `MirrorNodeEndpoint` and `RpcRelayEndpoint` are intentionally defined as distinct (currently empty) types rather than collapsing endpoint kind into a single enum. The protobuf specification defines these as separate messages so that type-specific fields can be added in future HIPs without restructuring the message. Preserving this pattern in the SDK means each endpoint type can independently gain fields (e.g., mirror node API version, RPC relay chain ID) without introducing breaking API changes. An enum-based approach would require bolting on new nullable companion types later, effectively retrofitting the oneOf pattern after the fact.
```
hendrikebbers: I understand the idea. My question would be like that:
if in any future fields would be added to that type, will there be meaningful defaults? The API without them could than stay in future and use the defaults. A more detailed api with additional params can live next to it.

hendrikebbers:
We should have a general discussion how we handle empty types like introduced at https://hips.hedera.com/hip/hip-113 - Maybe static factories are a good way:

```
@@oneOf(ipAddress, domainName)
@@oneOf(blockNode, mirrorNode, rpcRelay)
@@finalType
RegisteredServiceEndpoint {
    @@immutable @@nullable ipAddress: bytes
    @@immutable @@nullable domainName: string
    @@immutable port: uint32
    @@immutable @@default(false) requiresTls: bool
    @@immutable @@nullable blockNode: BlockNodeEndpoint
    @@immutable @@nullable mirrorNode: MirrorNodeEndpoint
    @@immutable @@nullable rpcRelay: RpcRelayEndpoint
}

RegisteredServiceEndpoint forBlockNode(......)
RegisteredServiceEndpoint forMirrorNode(.....)
RegisteredServiceEndpoint forRpcRelay(.....)
```

In the methods we can ignore the empty classes totally.

proposals/hips/hip-1137.md
```diff
@@ -0,0 +1,239 @@
...
+RegisteredNodeDeleteTransaction {
+    @@getSet registeredNodeId: uint64
+}
+```
+
+### RegisteredNodeUpdateTransaction
+
+Updates an existing registered node in the network address book. Must be signed by the registered node's current `adminKey`. If `adminKey` is being changed, both the old and new key must sign.
+
+When `serviceEndpoints` is set to a non-empty list, it replaces the existing endpoint list entirely. When left empty (the default), the existing endpoints are unchanged.
```
venilinvasilev: Can/Shall these endpoint be cleared? If yes, could we add a description how?

rwalworth:
> We should have a general discussion how we handle empty types like introduced at https://hips.hedera.com/hip/hip-113 - Maybe static factories are a good way:
> 
> ```
> @@oneOf(ipAddress, domainName)
> @@oneOf(blockNode, mirrorNode, rpcRelay)
> @@finalType
> RegisteredServiceEndpoint {
>     @@immutable @@nullable ipAddress: bytes
>     @@immutable @@nullable domainName: string
>     @@immutable port: uint32
>     @@immutable @@default(false) requiresTls: bool
>     @@immutable @@nullable blockNode: BlockNodeEndpoint
>     @@immutable @@nullable mirrorNode: MirrorNodeEndpoint
>     @@immutable @@nullable rpcRelay: RpcRelayEndpoint
> }
> 
> RegisteredServiceEndpoint forBlockNode(......)
> RegisteredServiceEndpoint forMirrorNode(.....)
> RegisteredServiceEndpoint forRpcRelay(.....)
> ```
> 
> In the methods we can ignore the empty classes totally.

Static factories would for sure give a cleaner API today, since `MirrorNodeEndpoint` and `RpcRelayEndpoint` are currently empty. Instead of constructing an empty marker type, users would just call `RegisteredServiceEndpoint.forMirrorNode(ip, 443, true)`. That's nice.

However, the factory approach could work against the same forward-compatibility argument we're making for keeping the marker types. If fields are added to `MirrorNodeEndpoint` or `RpcRelayEndpoint` in the future, the factory method signatures would need to evolve. Whether that's a problem depends on the language:

**Can add default parameter values (non-breaking):** C++, JavaScript/TypeScript, Python, Swift

**Cannot add default parameter values (breaking or workarounds needed):**
- **Java:** no default params; must add overloads (signature proliferation)
- **Go:** no default params and no overloading at all; must use functional options pattern, options struct, or new function names (e.g., `ForMirrorNodeWithApiVersion(...)` alongside `ForMirrorNode(...)`)
- **Rust:** no default params and no overloading; must use builder pattern or new function names

So for 3 of our 7 SDK languages, evolving a factory signature means either a breaking change or signature proliferation.

Maybe this then raises a bigger question: should individual SDKs have some freedom to add ergonomic conveniences (like factory methods) that make sense for their language and ecosystem, even if those aren't captured in the design doc? Or should the design doc prescribe everything and all SDKs implement exactly that? Maybe not everything can (or should) be captured in a cross-language design doc like this.

Either way, I think the marker types themselves should be in the design doc as the baseline API. Whether factories live alongside them could be a per-SDK decision, but I'd like to hear more thoughts on where that line should be.

proposals/hips/hip-1137.md
```diff
@@ -0,0 +1,239 @@
...
+}
+```
+
+### MirrorNodeEndpoint
+
+Describes a mirror node endpoint. Currently carries no additional data.
+
+```
+@@finalType
+MirrorNodeEndpoint {
```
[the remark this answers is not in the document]
hendrikebbers: Looked in the HIP and understand the idea behind it. We need a good definition how to handle that in SDK without the need for the user to create empty objects. I believe factory methods can be a really good way to handle that.

guides/api-guideline.md
hendrikebbers: Updated #140 based on your comments.

proposals/hips/hip-1137.md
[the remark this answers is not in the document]
rwalworth: I added Idea 2 to the doc - let me know what you think @hendrikebbers.

ivaylonikolov7 (commented):
When we implement the registered node APIs in the JS SDK, it would be great if **callers can easily see what kind of endpoint they have** when reading data (e.g. from `RegisteredNodeAddressBookQuery` or from a `RegisteredNode`'s `serviceEndpoints`).

Right now we have three endpoint subtypes that all extend `RegisteredServiceEndpoint` and share the same base fields (`ipAddress`/`domainName`, `port`, `requiresTls`). Only `BlockNodeServiceEndpoint` adds `endpointApi`. If we deserialize everything into a single shape, **we lose that type information** and users can't reliably tell block node vs mirror node vs RPC relay endpoints apart in code.

**Suggestion:** When deserializing from protobuf (which will use a `oneof` or equivalent for the three endpoint types) I'd recommend adding a **`kind`** field (e.g. `'block_node' | 'mirror_node' | 'rpc_relay'`) to each endpoint object, set during deserialization based on which variant was present. That gives users a simple `switch (endpoint.kind)` and keeps TypeScript narrowing working.

Without something like this, JS/TS users would get a list of endpoints that all look the same at runtime. Calling this out in the design doc or SDK implementation notes would help implementers handle it.

One last addition:  like I mentioned in DM conversation with Rob I will talk with the Mirror Node team so this new query is available in the browser via some kind of REST API.

venilinvasilev (dismissed):

PR #158, opened 2026-04-02
docs: typed receipts, universal records, and named response aliases for V3 transactions

rwalworth:
## Summary

This PR introduces per-transaction typed receipts and named response aliases to the V3 transaction API, while deliberately keeping records universal. It also adds `@@alias` to the meta-language and makes `Transaction` generic so the typed return value is preserved through the full build/sign/execute chain.

The core design positions are:

- **Typed receipts** — transactions that create entities (AccountCreate, TokenCreate, etc.) return a typed receipt subtype with only the fields relevant to that transaction, rather than a monolithic base type with 15 nullable fields.
- **Universal records** — records are intentionally not typed. The `TransactionRecord` protobuf evolves with cross-cutting protocol features that expand to new transaction types over time; named record subtypes would produce breaking changes as that set changes.
- **Named response aliases** — every transaction gets a named `@@alias` response type (e.g. `AccountCreateResponse`) for ergonomics at call sites, IDE completions, and documentation. These aliases carry no fields or methods.

> **Note:** The typed receipts and response aliases design carries real risk and should be validated with broader community input before V3 is finalized. See the design rationale section below.

---

## Motivation

The v2 `TransactionReceipt` contains roughly 15 optional/nullable fields — `accountId`, `fileId`, `contractId`, `topicId`, `tokenId`, `scheduleId`, `scheduledTransactionId`, `serials`, `topicSequenceNumber`, `topicRunningHash`, and more. Every field is nullable because it only applies to one specific transaction type. After creating an account, the user must know to read `.accountId` and accept that every other field is null. Static analysis cannot help, and documentation cannot fully substitute for type safety.

The v2 pattern also loses type information mid-chain. `buildAndExecute()` could theoretically return a typed value, but `build()` returns an untyped `Transaction` and `execute()` returns a bare `Response`, meaning the typed return value only existed at the first call site.

---

### 1. Generic `Transaction<$$Response>` — type preserved through signing chains

`Transaction` is now parameterized by `$$Response`. `build()` returns `Transaction<$$Response>`, `sign()` returns `Transaction<$$Response>`, and `execute()` returns `@@async $$Response`. The typed return value now flows through the entire chain:

```
// Before: type lost after build()
Transaction tx = new AccountCreateTransactionBuilder()...build(client);
Response response = tx.execute(client);              // untyped

// After: type preserved
Transaction<AccountCreateResponse> tx = new AccountCreateTransactionBuilder()...build(client);
AccountCreateResponse response = tx.execute(client); // typed
```

`fromBytes()` returns `Transaction<Response>` (the raw base) because the response type cannot be inferred from bytes alone. Callers that know the transaction type may cast explicitly.

### 2. Typed receipts for entity-creating transactions

Transactions that return entity-specific data in the receipt get a named receipt subtype with only the relevant fields. All other fields (status, exchangeRate) remain on the base `Receipt`.

| Transaction        | New receipt field(s)                      |
|--------------------|-------------------------------------------|
| AccountCreate      | `accountId`                               |
| FileCreate         | `fileId`                                  |
| ContractCreate     | `contractId`                              |
| TopicCreate        | `topicId`                                 |
| TokenCreate        | `tokenId`                                 |
| ScheduleCreate     | `scheduleId`, `scheduledTransactionId`    |
| TokenMint (NFT)    | `serials`, `totalSupply`                  |
| TopicMessageSubmit | `topicSequenceNumber`, `topicRunningHash` |

Transactions not in this table use `Receipt` directly — no typed subtype, no new classes.

### 3. Named response aliases via `@@alias`

Every transaction gets a named response alias for ergonomics:

```
@@alias AccountCreateResponse = Response<AccountCreateReceipt>  // entity-creating
@@alias AccountUpdateResponse = Response<Receipt>               // no receipt-specific output
```

These aliases must remain empty — no fields, no methods. All transaction-specific data lives on the typed receipt. The alias exists solely so that call sites, IDE completions, error messages, and documentation show a meaningful name rather than the raw generic expression.

`@@alias` is a new construct added to the meta-language (`guides/api-guideline.md`). Each language's best practice guide specifies the implementation pattern.

### 4. Universal records — intentionally not typed

`Response.queryRecord()` returns `Record<$$Receipt>` directly. There are no named record subtypes. The `$$Record` type parameter is removed from `Response` entirely, simplifying it to a single parameter:

```
// Before
Response<$$Receipt extends Receipt, $$Record extends Record<$$Receipt>>

// After
Response<$$Receipt extends Receipt>
```

### 5. `TransactionSupport` SPI — `$$Record` parameter removed

`TransactionSupport` previously had four type parameters including `$$Record`. Since records are never typed, this parameter is removed. The `convert(protoRecord)` method now returns `transactions.Record<$$Receipt>` directly.

### 6. `@@alias` added to meta-language

`guides/api-guideline.md` now defines the `@@alias` construct:

```
@@alias AliasName = TargetType
```

Rules: no body, may refer to any type expression including generic instantiations, must not be used to introduce new behavior. Implementation pattern is delegated to each language's best practice guide.

---

### Why typed receipts (and why it's still a risk)

The primary concern with per-type receipts is protocol evolution: if a future HIP adds a receipt field to an existing transaction type that currently returns only a base `Receipt`, the `queryReceipt()` return type changes — a breaking change across all SDK implementations.

The historical evidence from `transaction_receipt.proto` is encouraging but not conclusive:

- Every new receipt field introduced since genesis has been tied to a **new transaction type** (HCS topics, HTS tokens, scheduled transactions, NFTs, node management, registered nodes).
- The one cross-cutting addition — `block_number` (field 17, Mar 2026) — applies to **all** transactions and therefore lives on the base `Receipt` type. It does not require changes to any typed subtype.
- There is no historical precedent for an existing transaction type that currently returns a base `Receipt` gaining a transaction-specific receipt field after the fact.

This pattern holds because the receipt's purpose is narrow: "what entity was created, and did it succeed?" That question is answered when a transaction type is designed, not through later protocol amendments. The risk is estimated to be low but it is real, and the typed receipts approach should be revisited with broader community input before V3 ships.

### Why records are not typed

The `TransactionRecord` protobuf tells a different story. Its fields evolve with protocol features that cut across subsets of **existing** transaction types:

- `assessed_custom_fees` (field 13) — added for CryptoTransfer and token transfers (existing tx types)
- `paid_staking_rewards` (field 18) — applies to practically every transaction that adjusts stake
- `high_volume_pricing_multiplier` (field 23, added Feb 2026, HIP-1313) — applies to ~14 existing transaction types today, with the set expected to expand with future HIPs

If named record subtypes existed, each expansion of these cross-cutting fields to additional transaction types would change `queryRecord()`'s return type for each newly-affected transaction — a breaking change across all SDK implementations. The record is the protocol's extension surface by design; typing it creates a maintenance trap that gets worse over time.

### Why responses are named aliases

The protobuf `TransactionResponse` — the immediate gRPC reply from the node — has exactly two fields and zero structural changes in 7+ years of mainnet. It is a pre-consensus acknowledgement only; no transaction-specific data has ever appeared in it.

Named response aliases exist purely for ergonomics. `AccountCreateResponse` in an IDE completion or error message is immediately meaningful; `Response<AccountCreateReceipt>` requires the reader to decode the type to understand it. The risk is low given the proto history, but the same caution applied to typed receipts applies here.

---

## Files Changed

| File | Change |
|------|--------|
| `guides/api-guideline.md` | Added `@@alias` type alias construct to the meta-language |
| `v3-sandbox/prototype-api/transactions.md` | Generic `Transaction<$$Response>`, simplified `Response<$$Receipt>`, design rationale summary, `@@alias` patterns |
| `v3-sandbox/prototype-api/transactions-accounts.md` | `AccountCreateTransactionBuilder` with both type params, `AccountCreateReceipt`, `@@alias AccountCreateResponse` |
| `v3-sandbox/prototype-api/transactions-spi.md` | Removed `$$Record` type parameter from `TransactionSupport` |

---

## Breaking Changes

**None** — this is a V3 prototype API definition. No existing SDK implementation is affected.

---

## Open Questions

These should be discussed before V3 is finalized:

1. **Typed receipts risk** — the historical evidence is encouraging but a single protocol change adding a receipt field to an existing non-entity-creating transaction would require breaking changes across all SDKs. Is the community comfortable with this trade-off, or should V3 stay with a universal `Receipt` and accept the v2 nullable-field DX problem?

2. **Which transactions need typed receipts** — the table in this PR is a working draft based on the current `transaction_receipt.proto` fields. It should be reviewed by SDK maintainers and the consensus node team to confirm completeness and accuracy.

3. **`@@alias` implementation pattern** — language-specific best practice guides need to define how `@@alias` maps to each language. For languages without native alias support, the fallback pattern must be specified.

### DIFF
--- a/guides/api-guideline.md
+++ b/guides/api-guideline.md
@@ -199,33 +199,6 @@ abstraction FruitFactory<$$Product extends Fruit> {
 In the given example, the `FruitFactory` provides a `create` method that returns a `Fruit`.
 `Fruit` must be a concrete complexe type (not a generic type parameter).
 
-#### Type Aliases
-
-A type alias gives a name to an existing type expression without creating a new, distinct type. Aliases
-are useful for naming specific instantiations of generic types so that the name appears in call sites,
-IDE completions, and documentation instead of the raw generic expression.
-
-Use the following syntax to declare a type alias:
-
-```
-@@alias AliasName = TargetType
-```
-
-A concrete example:
-
-```
-@@alias StringList = list<string>
-@@alias AccountCreateResponse = Response<AccountCreateReceipt>
-```
-
-Rules and guidelines for `@@alias`:
-
-- An alias declaration has no body — it cannot declare fields or methods.
-- An alias may refer to any type expression, including generic instantiations.
-- Aliases must not be used to introduce new behavior. If any field or method is needed, define a new
-  complex type using `extends` instead.
-- Language implementations translate `@@alias` according to their language-specific best practice guide.
-
 ### Enumerations
 
 Enumerations can be defined using the following syntax:
@@ -259,6 +232,33 @@ enum EnumName {
     int calcPrimitiveType()
 }
 ```
+
+### Type Aliases
+
+A type alias gives a name to an existing type expression without creating a new, distinct type. Aliases
+are useful for naming specific instantiations of generic types so that the name appears in call sites,
+IDE completions, and documentation instead of the raw generic expression.
+
+Use the following syntax to declare a type alias:
+
+```
+@@alias AliasName = TargetType
+```
+
+A concrete example:
+
+```
+@@alias StringList = list<string>
+@@alias AccountCreateResponse = Response<AccountCreateReceipt>
+```
+
+Rules and guidelines for `@@alias`:
+
+- An alias declaration has no body — it cannot declare fields or methods.
+- An alias may refer to any type expression, including generic instantiations.
+- Aliases must not be used to introduce new behavior. If any field or method is needed, define a new
+  complex type using `extends` instead.
+- Language implementations translate `@@alias` according to their language-specific best practice guide.
 
 ### Attributes
 
--- a/guides/api-guideline.md
+++ b/guides/api-guideline.md
@@ -233,33 +233,6 @@ enum EnumName {
 }
 ```
 
-### Type Aliases
-
-A type alias gives a name to an existing type expression without creating a new, distinct type. Aliases
-are useful for naming specific instantiations of generic types so that the name appears in call sites,
-IDE completions, and documentation instead of the raw generic expression.
-
-Use the following syntax to declare a type alias:
-
-```
-@@alias AliasName = TargetType
-```
-
-A concrete example:
-
-```
-@@alias StringList = list<string>
-@@alias AccountCreateResponse = Response<AccountCreateReceipt>
-```
-
-Rules and guidelines for `@@alias`:
-
-- An alias declaration has no body — it cannot declare fields or methods.
-- An alias may refer to any type expression, including generic instantiations.
-- Aliases must not be used to introduce new behavior. If any field or method is needed, define a new
-  complex type using `extends` instead.
-- Language implementations translate `@@alias` according to their language-specific best practice guide.
-
 ### Attributes
 
 Attributes can be defined using the following syntax:
--- a/guides/api-guideline.md
+++ b/guides/api-guideline.md
@@ -199,6 +199,36 @@ abstraction FruitFactory<$$Product extends Fruit> {
 In the given example, the `FruitFactory` provides a `create` method that returns a `Fruit`.
 `Fruit` must be a concrete complexe type (not a generic type parameter).
 
+#### Type Aliases
+
+A type alias gives a name to an existing type expression without creating a new, distinct type. Aliases
+are useful for naming specific instantiations of generic types so that the name appears in call sites,
+IDE completions, and documentation instead of the raw generic expression.
+
+Use the following syntax to declare a type alias:
+
+```
+@@alias AliasName = TargetType
+```
+
+A concrete example:
+
+```
+@@alias StringList = list<string>
+@@alias AccountCreateResponse = Response<AccountCreateReceipt>
+```
+
+Rules and guidelines for `@@alias`:
+
+- An alias declaration has no body — it cannot declare fields or methods.
+- An alias may refer to any type expression, including generic instantiations.
+- Aliases must not be used to introduce new behavior. If any field or method is needed, define a new
+  complex type using `extends` instead.
+- Language implementations translate `@@alias` to the most idiomatic alias mechanism available
+  (e.g. `type` in TypeScript/Rust/Go, `typealias` in Swift, `using` in C++). Languages without a
+  native alias mechanism (e.g. Java) must implement the alias as an empty `final` class or record
+  that extends the target type. The language-specific best practice guide specifies the exact pattern.
+
 ### Enumerations
 
 Enumerations can be defined using the following syntax:
--- a/guides/api-guideline.md
+++ b/guides/api-guideline.md
@@ -224,10 +224,7 @@ Rules and guidelines for `@@alias`:
 - An alias may refer to any type expression, including generic instantiations.
 - Aliases must not be used to introduce new behavior. If any field or method is needed, define a new
   complex type using `extends` instead.
-- Language implementations translate `@@alias` to the most idiomatic alias mechanism available
-  (e.g. `type` in TypeScript/Rust/Go, `typealias` in Swift, `using` in C++). Languages without a
-  native alias mechanism (e.g. Java) must implement the alias as an empty `final` class or record
-  that extends the target type. The language-specific best practice guide specifies the exact pattern.
+- Language implementations translate `@@alias` according to their language-specific best practice guide.
 
 ### Enumerations
 
--- a/v3-sandbox/prototype-api/transactions.md
+++ b/v3-sandbox/prototype-api/transactions.md
@@ -237,38 +237,38 @@ specific transaction type. After creating an account, for example, the user must
 and accept that all other fields are null. The compiler cannot help, and documentation cannot fully
 substitute for type safety.
 
-### Why receipts are typed (risk: ~5%)
+### Why receipts are typed
 
-The concern with per-type receipts is that a future protocol change could add a field to an existing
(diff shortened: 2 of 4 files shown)


v3-sandbox/prototype-api/transactions.md
```diff
@@ -62,69 +66,91 @@ abstraction TransactionBuilder<$$Transaction extends TransactionBuilder, $$Respo
...
-  // Deserialize a transaction from bytes
+  bytes toBytes()
+
+  // Deserialize a transaction from bytes. Returns a raw Transaction<Response> because the response
+  // type cannot be inferred from bytes alone. If the transaction type is known at the call site
+  // (e.g. in a multi-party signing flow), language implementations may provide typed overloads
+  // or allow an explicit cast to the expected Transaction<$$Response>.
   @@static
-  Transaction fromBytes(transactionBytes: bytes)
+  Transaction<Response> fromBytes(transactionBytes: bytes)
```
ddeskov-limechain: Maybe change `fromBytes()` to return `Transaction<Response<Receipt>>` :

```
@@static
Transaction<Response<Receipt>> fromBytes(transactionBytes: bytes)
```

Before this PR, `fromBytes()` returned non-generic `Transaction` — and it was correct because `Transaction` itself wasn't generic. Now `Transaction` is generic but returns `Transaction<Response>` where `Response` is a raw type (`Response` requires a type parameter: `Response<$$Receipt extends Receipt>`).

ddeskov-limechain (commented):
I have some suggestions based on the changes introduced in the PR

v3-sandbox/prototype-api/transactions.md
rwalworth: Agreed - pushed up a fix. Changed `Transaction<Response>` to `Transaction<Response<Receipt>>` in the `fromBytes()` declaration, its inline comment, and the multi-party signing example comment.

v3-sandbox/prototype-api/transactions.md
```diff
@@ -151,34 +178,40 @@ FooReceipt receipt = response.queryReceipt();
...
 
-For flows where a dApp builds a transaction without a transactionId or nodeAccountIds and sends it to a wallet for completion.
+For flows where a dApp builds a transaction without a transactionId or nodeAccountIds and sends it to a wallet for
+completion. Both sides treat the transaction as raw (`Transaction<Response>`) because the wallet operates on
+arbitrary transaction types.
 
 ```
 // dApp builds without a client — no transactionId or nodeAccountIds generated
-Transaction tx = new TransferTransactionBuilder()
+Transaction<Response> tx = new TransferTransactionBuilder()
```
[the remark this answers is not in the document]
rwalworth: Good catch. Since `TransferTransaction` doesn't produce any entity-specific receipt fields, `TransferTransactionBuilder` doesn't need its own receipt subtype or named response alias — it should use `Response<Receipt>` directly as its `$$Response`. That makes `build()` return `Transaction<Response<Receipt>>` and the HIP-745 example is type-correct in all languages without a cast. Pushed up a comment in the example to make this explicit.

guides/api-guideline.md
```diff
@@ -199,6 +199,33 @@ abstraction FruitFactory<$$Product extends Fruit> {
...
+@@alias AccountCreateResponse = Response<AccountCreateReceipt>
+```
+
+Rules and guidelines for `@@alias`:
+
+- An alias declaration has no body — it cannot declare fields or methods.
+- An alias may refer to any type expression, including generic instantiations.
+- Aliases must not be used to introduce new behavior. If any field or method is needed, define a new
+  complex type using `extends` instead.
+- Language implementations translate `@@alias` according to their language-specific best practice guide.
```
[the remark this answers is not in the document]
hendrikebbers: I agree with @ddeskov-limechain - My assumption is that we will have some `X<Y<Z>>` code blocks in the API that first look complete but will be very easy to use for the enduser. A SDK developer needs to understand it and needs to implement it but for the usage it is always quite easy.

If a language proved a good feature to define alias to make the public API look more easy it will be totally fine to use that (like use Records in Java to create immutable types) but it should not be a base definition in the meta-language.

The pseudo-api created by the meta language should be very concrete (`X<Y<Z>>`) because it defines "How it must be". An alias is more an "What we can do for make it more easy". Great to do it in SDKs by using language specific features but not in meta spec from my point of view.

hendrikebbers (commented):
M suggestion is to create everything with generics (maybe even in a complete looking way) first. Once we have that (a solid technical base) we can discuss how we can make it more easy. This PR tries to mix both points and therefore will end in many discussions....

v3-sandbox/prototype-api/transactions.md
```diff
@@ -188,11 +221,67 @@ bytes txBytes = tx.toBytes();
...
 
+## Design Rationale
+
+Receipts are typed per-transaction because the receipt's purpose is narrow ("what entity was created?")
+and 7 years of mainnet history show that receipt fields have always been introduced alongside new
+transaction types, never added to existing ones after the fact. Records are intentionally not typed
+because the `TransactionRecord` protobuf evolves with cross-cutting protocol features that expand to new
+transaction types over time — named record subtypes would produce breaking changes as that set grows.
+
+**This is a working draft.** Typed receipts and responses carry real protocol-evolution risk and these
```
hendrikebbers: I agree that we need to learn more here. My current point: Only because protobuf does it we do not need to do it if we can provide a more simple/smaller/concreter scope/api to the app developer. Nothing we need to decide today :)

hendrikebbers (approved):

PR #165, opened 2026-04-09
docs: add guidance for immutable byte arrays in Java

### LINKED ISSUE #161: Immutable byte arrays in Java
The meta-language defines `@@immutable id: bytes`, but Java arrays are mutable references and a Java `record` would expose the internal `byte[]` reference unmodifiably. There is no guidance in the best-practices guide on how to model immutable byte arrays.

**Prototype handling:** Implemented `Ledger` as a non-record `final class` with a defensive copy on construction and on every getter call. Custom `equals` / `hashCode` / `toString` are required.

**Recommendation:** Add a section to api-best-practices-java about "Immutable byte arrays" and what pattern to use (defensive copy class vs. `ByteBuffer.asReadOnlyBuffer()` vs. a custom `ImmutableBytes` wrapper).

Nesar976:
Add guidance for handling immutable byte arrays in Java using defensive copies and document alternative approaches.

Fixes #161

### DIFF
--- a/guides/api-best-practices-java.md
+++ b/guides/api-best-practices-java.md
@@ -290,6 +290,68 @@ public class Person {
 }
 ```
 
+### Immutable byte arrays
+
+#### The problem
+
+Java's `byte[]` is mutable. Any code holding a reference to the array can modify its contents, breaking the guarantee
+of `@@immutable id: bytes` from the meta-language.
+
+A Java `record` does not solve this – it stores the reference to the array, not a defensive copy:
+
+```java
+byte[] raw = new byte[]{1, 2, 3};
+Ledger ledger = new Ledger(raw);
+raw[0] = 99; 
+```
+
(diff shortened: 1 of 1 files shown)


guides/api-best-practices-java.md
```diff
@@ -290,6 +290,68 @@ public class Person {
...
+    public byte[] getId() {
+        return Arrays.copyOf(id, id.length);
+    }
+}
+```
+
+Implement `equals` using `Arrays.equals`, `hashCode` using `Arrays.hashCode`, and `toString` using `Arrays.toString`.
+The default `Object` implementations compare references, not contents, and produce unusable output for arrays.
+
+#### Alternative approaches
```
hendrikebbers, line 336: We should remove that section. Initial we do not want alternatives. In best we work with copies here

PR #179, opened 2026-04-10
docs: fix HIP-1261 design doc issues

rwalworth:
## Summary

This PR corrects several inaccuracies in the HIP-1261 (`FeeEstimateQuery`) SDK design document. The issues range from an incorrect transport assumption (gRPC vs REST) to missing field clarifications and a typo.

**Key Changes:**
- Correct the transport from gRPC to mirror node REST (`POST /api/v1/network/fees`)
- Remove `FeeEstimateQuery extends Query<FeeEstimateResponse>` — replace with a plain class declaration that directs SDK authors to the mirror node REST query pattern
- Replace gRPC status codes in the retry policy with HTTP equivalents
- Remove `notes` and `mode` fields from `FeeEstimateResponse` (neither is returned by the mirror node)
- Align `FeeEstimateResponse` field names with the actual mirror node JSON keys
- Fix `TopicMessageQuery` → `TopicMessageSubmitTransaction` typo

### Transport: gRPC → mirror node REST

The introduction incorrectly described the feature as backed by a "Mirror gRPC `NetworkService.getFeeEstimate` endpoint." The actual transport is `POST /api/v1/network/fees` (REST), accepting a protobuf-binary body and returning JSON.

- Updated the introduction paragraph to reference the REST endpoint
- Changed `FeeEstimateQuery extends Query<FeeEstimateResponse>` to a plain class declaration with a note explaining it communicates via the mirror node REST API and should be modeled after the SDK's existing mirror node REST query class (e.g. `MirrorNodeContractQuery`)

### Retry policy: gRPC status codes → HTTP equivalents

The retry policy named `UNAVAILABLE`, `DEADLINE_EXCEEDED`, and `INVALID_ARGUMENT` — all gRPC status codes — which would be interpreted inconsistently across SDK implementations using REST.

- `UNAVAILABLE` → HTTP 503
- `DEADLINE_EXCEEDED` → request timeout (`URLError.timedOut`, `SocketTimeoutException`, etc.)
- `INVALID_ARGUMENT` → HTTP 400 (no retry)

Updated the three corresponding test plan entries to match.

### Removed `notes` from `FeeEstimateResponse`

The mirror node `FeeEstimateResponse` schema (`openapi.yml`) defines exactly four fields: `network`, `node`, `service`, `total`. There is no `notes` field. Removed the field from the type definition and the example code; it can be re-added when the mirror node implements it.

### Removed `mode` from `FeeEstimateResponse`

The mirror node JSON response does not include a `mode` field — it is only a request query parameter. Since the SDK can't know which mode the mirror node actually used (e.g. STATE silently falling back to INTRINSIC), echoing the requested mode back would be potentially misleading and adds no value. Removed from the type definition, example code, and test plan.

### Field name alignment

The `FeeEstimateResponse` type definition used `networkFee`, `nodeFee`, and `serviceFee`, while the mirror node JSON uses `network`, `node`, and `service`. The example code was already using the correct JSON names. Renamed the fields in the type definition to match and added `// JSON key:` comments for clarity.

### `TopicMessageQuery` typo

Line 145 listed `TopicMessageQuery` as an example of a chunked transaction. `TopicMessageQuery` is a topic subscription, not a transaction. Fixed to `TopicMessageSubmitTransaction`, which already appeared correctly in the bullet list immediately below.

## Files Changed Summary

| File | Change |
|------|--------|
| `proposals/hips/hip-1261.md` | All fixes above |

## Breaking Changes

**None.** This PR only corrects the design document. No SDK source code is modified.

### DIFF
--- a/proposals/hips/hip-1261.md
+++ b/proposals/hips/hip-1261.md
@@ -1,6 +1,6 @@
 # Fee Estimation Query API
 
-This design introduces a **fee estimation query** that allows users, SDKs, and tools to query expected transaction fees without submitting transactions to the network. The feature is backed by a new Mirror gRPC `NetworkService.getFeeEstimate` endpoint that accepts a `FeeEstimateQuery` with an `EstimateMode` and returns a structured `FeeEstimateResponse` containing network, node, and service fee components.
+This design introduces a **fee estimation query** that allows users, SDKs, and tools to query expected transaction fees without submitting transactions to the network. The feature is backed by the mirror node REST endpoint `POST /api/v1/network/fees`, which accepts a serialized HAPI transaction (protobuf binary) with an `EstimateMode` query parameter and returns a `FeeEstimateResponse` JSON body containing network, node, and service fee components.
 
 ## Summary
 
@@ -82,23 +82,20 @@ NetworkFee {
 ```
 // The response containing the estimated transaction fees.
 FeeEstimateResponse {
-    // The mode that was used to calculate the fees.
-    @immutable mode: FeeEstimateMode
-
     // The network fee component which covers the cost of gossip, consensus,
     // signature verifications, fee payment, and storage.
-    @immutable networkFee: NetworkFee
+    // JSON key: "network"
+    @immutable network: NetworkFee
 
     // The node fee component which is to be paid to the node that submitted the
     // transaction to the network.
-    @immutable nodeFee: FeeEstimate
+    // JSON key: "node"
+    @immutable node: FeeEstimate
 
     // The service fee component which covers execution costs, state saved in the
     // Merkle tree, and additional costs to the blockchain storage.
-    @immutable serviceFee: FeeEstimate
-
-    // An array of strings for any caveats (e.g., ["Fallback to worst-case due to missing state"]).
-    @immutable notes: list<string>
+    // JSON key: "service"
+    @immutable service: FeeEstimate
 
     // The sum of the network, node, and service subtotals in tinycents.
     @immutable total: uint64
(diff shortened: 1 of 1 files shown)


Dosik13 (approved):

ivaylonikolov7 (dismissed):

proposals/hips/hip-1261.md
Dosik13, line 24: based on this information here https://github.com/hiero-ledger/hiero-mirror-node/blob/main/rest/api/v1/openapi.yml#L3456, I think we should change the default value to be `INTRINSIC`

gsstoykov (dismissed):

ivaylonikolov7 (dismissed):

proposals/hips/hip-1261.md
```diff
@@ -32,15 +32,16 @@ FeeExtra {
...
-    @immutable count: uint32
+    @immutable count: uint64
 
     // The charged count of items as calculated by max(0, count - included).
-    @immutable charged: uint32
+    @immutable charged: uint64
 
     // The fee price per unit in tinycents.
+    // JSON key: "fee_per_unit"
     @immutable feePerUnit: uint64
```
Dosik13, line 45: Why in some places there is JSON key and on others there isn't. Is there a rule or something like this I don't know about :D

gsstoykov (approved):

proposals/hips/hip-1261.md
rwalworth: There's no established repo standard for this - I just added the comments for clarity. The pattern I followed:

1. **Field name ≠ JSON key** - when the camelCase field name doesn't trivially map to the actual snake_case JSON key:
   - `feePerUnit` → `"fee_per_unit"`
   - `highVolumeMultiplier` → `"high_volume_multiplier"`

2. **Rename disambiguation** - `network`, `node`, and `service` in `FeeEstimateResponse` technically have the same field name as their JSON key, so by rule 1 they don't strictly need comments. But they were renamed in this PR from `networkFee`, `nodeFee`, `serviceFee`, so I added comments to make it explicit.

Fields like `name`, `included`, `count`, `charged`, `subtotal`, `base`, `extras`, `multiplier`, and `total` are identical in both the type definition and the mirror node JSON response, so no comment needed there.

That said, it would be worth establishing a repo-wide standard for this going forward - happy to partake in any discussion there.

PR #185, opened 2026-04-12
docs(cpp): expand API best-practices parity 

### LINKED ISSUE #183: [V3 Docs] Expand C++ API best-practices guide to parity baseline with C++17 fallback mapping
## Summary

The C++ best-practices guide currently covers only a minimal type mapping and lacks implementation-critical sections for ownership, errors, async behavior, nullability, and thread safety.

This issue proposes a full parity expansion with C++17 as the primary standard target for SDK guidance. C++20 mappings may be documented only as optional alternatives where useful.

## Proposal

Expand the C++ guide with these sections and examples:

1. Type mapping table with C++17 as primary mapping and optional C++20 alternatives
2. Error model mapping for `@@throws`
3. Async model mapping for `@@async`
4. Factory and static method conventions
5. Nullability and collection semantics
6. Immutability and thread-safety guidance (including ownership and const-correctness)
7. One runnable build-sign-send example

For C++20-only chrono/timezone types, provide clear C++17-first guidance and list C++20 forms as optional.

## Why this matters

- Aligns docs with the actual C++ SDK standard target (C++17)
- Reduces contributor confusion and review churn
- Improves implementation-readiness with concrete ownership/concurrency patterns
- Maintains parity with the language-guide baseline from #180

## Acceptance Criteria

- C++ guide includes all 7 baseline sections
- Type mapping table treats C++17 as the primary/default mapping target
- C++20 mappings are clearly marked as optional alternatives, not preferred defaults
- `@@throws` guidance defines stable SDK-level error modeling strategy
- `@@async` guidance provides a clear future-based contract and sync/async consistency notes
- Nullability guidance uses explicit optional semantics and non-null collection behavior
- Immutability/thread-safety guidance includes ownership and synchronization patterns
- Runnable build-sign-send example is included and self-contained

## Out of Scope

- Enforcing compiler flags in SDK repositories
- Full migration planning for all C++ runtime/toolchain variants

Omswastik-11:
Closes #183

### DIFF
--- a/guides/api-best-practices-cpp.md
+++ b/guides/api-best-practices-cpp.md
@@ -23,24 +23,50 @@ Where C++20 offers a clearer standard type, that mapping is listed as an optiona
 | `map<KEY, VALUE>` | `std::map<KEY, VALUE>`                        | `std::map<KEY, VALUE>`                            | Use `std::unordered_map` if needed      |
 | `type`            | `std::type_index`                             | `std::type_index`                                 | From `<typeindex>`                      |
 | `uuid`            | UUID library type                             | UUID library type                                 | No standard UUID until C++26            |
-| `date`            | Custom date struct or library type            | `std::chrono::year_month_day`                     | `year_month_day` is C++20               |
+| `date`            | `date::year_month_day`                        | `std::chrono::year_month_day`                     | C++17: Howard Hinnant date library      |
 | `time`            | Custom time struct or library type            | `std::chrono::hh_mm_ss<std::chrono::nanoseconds>` | `hh_mm_ss` is C++20                     |
 | `dateTime`        | `std::chrono::system_clock::time_point`       | `std::chrono::system_clock::time_point`           | -                                       |
-| `zonedDateTime`   | `time_point` plus explicit IANA zone string   | `std::chrono::zoned_time`                         | `zoned_time` is C++20                   |
-| `function<...>`   | `std::function<...>`                          | `std::function<...>`                              | Prefer templates for hot paths          |
+| `zonedDateTime`   | `date::zoned_time<std::chrono::system_clock::duration>` | `std::chrono::zoned_time`                | C++17: Howard Hinnant date library      |
+| `function<...>`   | `std::function<...>`                          | `std::function<...>`                              | Use templates for one-shot invocation; use `std::function` when callables are stored |
 
 ## 2. Error Model Mapping For @@throws
 
 `@@throws(error-type-a[, ...])` maps to typed C++ error handling at the SDK boundary.
 
+Recommended V3 pattern: use a discriminated result type as the primary model.
+
+- Return `HieroResult<T>` for fallible methods instead of throwing by default.
+- Use one stable `HieroError` shape (for example, `code` plus optional message/details).
+- Multiple `@@throws(...)` entries still map to the same `HieroResult<T>`; the error code distinguishes cases.
+- For methods without a value return, use a status result (for example, `HieroStatus`) that carries success/failure.
+
+Mapping sketch:
+
+```cpp
+// Meta-language:
+// @@throws(not-found-error, invalid-argument-error)
+// AccountBalance getBalance(id: AccountId)
+
+// C++17 translation:
+HieroResult<AccountBalance> getBalance(const AccountId& id);
+
+auto result = client.getBalance(id);
+if (!result) {
+    log(result.error().message);
+    return;
+}
+auto& balance = *result;
+```
+
 Recommended pattern:
 
 - Public APIs should expose stable SDK-level error identifiers, not transport-specific status enums.
-- Prefer a structured error object with `code` and `message` when avoiding exceptions.
-- If exceptions are enabled in the SDK, use a base exception type and specific subclasses.
+- Result types are the recommended V3 default in C++17.
+- The current V2 SDK uses an exception hierarchy; document this for contributor context.
+- Exception hierarchies are still a valid fallback when a project explicitly relies on exceptions.
 - Preserve lower-level causes in logs or nested error payloads.
 
-Example exception pattern:
+Example exception fallback pattern:
 
 ```cpp
 #include <stdexcept>
@@ -67,27 +93,44 @@ public:
 
 `@@async` should map to non-blocking APIs that return futures or completion objects.
 
+Recommended V3 pattern: an explicit client-owned executor or thread pool.
+
 Recommended pattern:
 
-- Use `std::future<T>` for simple async contracts.
-- For richer cancellation or continuations, use a runtime-specific async abstraction.
+- Submit work items to an explicit thread pool or executor owned by the client/runtime.
+- Return a future-like handle from the pooled submission.
 - If both sync and async variants exist, keep shared execution logic in one internal implementation.
 - Do not mix callback and future styles in a single public method signature.
 
-Example:
+`std::async(std::launch::async, ...)` note:
+
+- V2-style `std::async` is acceptable for simple, infrequent one-off operations.
+- At SDK scale, prefer pooled executors because `std::async` can create one OS thread per call,
+  futures are not composable in C++17, and future destruction may block if `.get()` is never called.
+
+Example executor-oriented sketch:
 
 ```cpp
 #include <future>
-#include <vector>
-#include <cstdint>
+#include <memory>
+#include <utility>
+
+class Executor {
+public:
+    template <typename Fn>
+    auto submit(Fn&& fn) -> std::future<decltype(fn())>;
+};
 
 class Client {
 public:
-    std::future<std::vector<std::uint8_t>> submitAsync(std::vector<std::uint8_t> txBytes) const {
-        return std::async(std::launch::async, [payload = std::move(txBytes)]() {
-            return payload;
-        });
+    explicit Client(std::shared_ptr<Executor> executor) : executor_(std::move(executor)) {}
+
+    std::future<int> submitAsync(int request) const {
+        return executor_->submit([req = request]() { return req + 1; });
     }
+
+private:
+    std::shared_ptr<Executor> executor_;
 };
 ```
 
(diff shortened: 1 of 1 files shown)


guides/api-best-practices-cpp.md
```diff
@@ -1,24 +1,287 @@
...
+| `type`            | `std::type_index`                             | `std::type_index`                                 | From `<typeindex>`                      |
+| `uuid`            | UUID library type                             | UUID library type                                 | No standard UUID until C++26            |
+| `date`            | Custom date struct or library type            | `std::chrono::year_month_day`                     | `year_month_day` is C++20               |
+| `time`            | Custom time struct or library type            | `std::chrono::hh_mm_ss<std::chrono::nanoseconds>` | `hh_mm_ss` is C++20                     |
+| `dateTime`        | `std::chrono::system_clock::time_point`       | `std::chrono::system_clock::time_point`           | -                                       |
+| `zonedDateTime`   | `time_point` plus explicit IANA zone string   | `std::chrono::zoned_time`                         | `zoned_time` is C++20                   |
+| `function<...>`   | `std::function<...>`                          | `std::function<...>`                              | Prefer templates for hot paths          |
+
+## 2. Error Model Mapping For @@throws
```
rwalworth, line 32: **suggestion (blocking):** The error section only covers the exception hierarchy pattern. That's what the current C++ SDK uses (V2 throws `PrecheckStatusException`, `ReceiptStatusException`, etc., all inheriting from `std::exception`) - but exception-based error handling is not the dominant modern C++17 pattern for SDK-level APIs.

Major C++17 SDK codebases (Abseil/gRPC, LLVM, Folly) use a discriminated result type instead: a value on success, a structured error on failure, where the caller is forced to check before accessing the value. Exception hierarchies don't enforce that - a caller can silently swallow an exception or never check. Result types also work in `-fno-exceptions` builds.

**How `@@throws` maps to a result type**

`@@throws(error-type-a[, ...])` declares that a method can fail. In a result type model, the return type becomes the discriminated union - the annotation drives what the error type needs to carry:

```cpp
// Meta-language:
//   @@throws(not-found-error, invalid-argument-error)
//   AccountBalance getBalance(id: AccountId)

// C++17 result type translation:
// HieroResult<T> is a discriminated union — either T or a HieroError
HieroResult<AccountBalance> getBalance(const AccountId& id);

// Caller must inspect the result before accessing the value:
auto result = client.getBalance(id);
if (!result) {
    // result.error() carries the status code and message
    log(result.error().message);
    return;
}
auto& balance = *result;  // safe — success path confirmed
```

The error type holds a status code and an optional message. Multiple `@@throws` entries still produce a single `HieroResult<T>` - all failure cases reduce to a status code, so no union of error types is needed. For void-returning methods, `HieroStatus` (result of a `void` success or a `HieroError`) serves the same role.

This section should:
1. Lead with the result type pattern as the recommended V3 approach, with the mapping sketch above
2. Note that the current SDK uses exceptions (so contributors reading V2 aren't confused)
3. Mention the exception hierarchy as a valid fallback when a project explicitly enables and relies on exceptions

guides/api-best-practices-cpp.md
```diff
@@ -1,24 +1,287 @@
...
+    std::string code_;
+};
+
+class NotFoundError : public HieroError {
+public:
+    NotFoundError() : HieroError("not-found-error", "Resource not found") {}
+};
+```
+
+## 3. Async Model Mapping For @@async
```
rwalworth, line 92: **issue (blocking):** Section 3 recommends `std::async(std::launch::async, ...)` as the primary async pattern, and Section 7 uses it too. This is what the current C++ SDK does (e.g. `Executable::executeAsync()`) - but it's a V2 limitation, not a practice to carry forward.

The problems with `std::async(std::launch::async, ...)` at SDK scale:
- Spawns a new OS thread per call (~0.5 ms overhead, limited thread counts)
- `std::future`'s destructor blocks if the future came from `std::async` - a silent performance trap
- Not composable in C++17 (no `.then()` continuations)

This is explicitly documented as problematic by Folly, Qt6, and Boost.Asio teams, all of which use explicit thread pool/executor models instead. The recommended C++17 pattern for SDK async operations is an executor-owned thread pool where tasks are submitted as work items rather than spawning threads per call.

Please restructure this section to:
1. Recommend an explicit thread pool as the V3 async model (e.g., a `Client`-owned executor)
2. Note that `std::async(std::launch::async)` is what V2 uses and call out its limitations: one new OS thread per call with no pooling or reuse, non-composable futures in C++17, and a blocking destructor if the caller drops the future without calling `.get()`
3. Keep `std::async` as an acceptable fallback for simple, infrequent one-off operations - it's a correctness concern only at scale.

The Section 7 example can stay as-is (standalone illustration), but the guidance text in Section 3 should lead with the thread pool recommendation.

guides/api-best-practices-cpp.md
```diff
@@ -1,24 +1,287 @@
...
+| `type`            | `std::type_index`                             | `std::type_index`                                 | From `<typeindex>`                      |
+| `uuid`            | UUID library type                             | UUID library type                                 | No standard UUID until C++26            |
+| `date`            | Custom date struct or library type            | `std::chrono::year_month_day`                     | `year_month_day` is C++20               |
+| `time`            | Custom time struct or library type            | `std::chrono::hh_mm_ss<std::chrono::nanoseconds>` | `hh_mm_ss` is C++20                     |
+| `dateTime`        | `std::chrono::system_clock::time_point`       | `std::chrono::system_clock::time_point`           | -                                       |
+| `zonedDateTime`   | `time_point` plus explicit IANA zone string   | `std::chrono::zoned_time`                         | `zoned_time` is C++20                   |
+| `function<...>`   | `std::function<...>`                          | `std::function<...>`                              | Prefer templates for hot paths          |
```
rwalworth: **suggestion (non-blocking):** The type mapping note for `function<...>` says "Prefer templates for hot paths" - this is correct but undersells the distinction. The actual rule used by major C++17 libraries (LLVM, range-v3, Abseil) is:

- **Template parameter**: when the callable is invoked within the same function (enables full inlining, zero overhead)
- **`std::function`**: only when the callable must be *stored* (class member, container, async continuation) and its type is not known at compile time

The current SDK uses `std::function` everywhere in public APIs - for stored signers, listeners, and async callbacks - which is correct for those cases. But it also uses it for one-shot callbacks that could be templates. Worth documenting the ownership distinction clearly, not just "prefer templates for hot paths."

guides/api-best-practices-cpp.md
```diff
@@ -1,24 +1,287 @@
...
+struct TransactionMemo {
+    std::optional<std::string> text;
+};
+
+struct ValidationResult {
+    std::vector<std::string> warnings;
+};
+```
+
+## 6. Immutability And Thread-Safety Guidance
```
rwalworth, line 193: **issue (blocking):** Section 6 covers `std::lock_guard` for simple mutation protection, which is correct. But there are two common C++17 thread patterns missing:

**1. `std::unique_lock` + `std::condition_variable`**

`std::lock_guard` cannot be used with `std::condition_variable` - `std::unique_lock` is required. This is how the SDK's `Client` actually coordinates its background network update thread:

```cpp
std::mutex mMutex;
std::condition_variable mConditionVariable;

// Usage:
std::unique_lock<std::mutex> lock(mImpl->mMutex);
mImpl->mConditionVariable.wait(lock, [this] { /* condition */ });
```

Any V3 component that needs to wait on a background condition (not just protect a mutation) needs this pattern.

**2. `std::atomic<T>` for counters and flags**

The SDK uses `std::atomic<int64_t>` in `TransactionId` for a lock-free nonce counter. `std::atomic<bool>` for shutdown flags and `std::atomic<int>` for simple counters are standard C++17 and worth documenting - though a note that complex lock-free patterns should be avoided (Abseil's guidance is explicit: raw atomics are nearly untestable under concurrency) is warranted.

Please add both patterns to this section.

guides/api-best-practices-cpp.md
```diff
@@ -1,24 +1,287 @@
...
+public:
+    std::uint64_t next() {
+        std::lock_guard<std::mutex> guard(mutex_);
+        return ++value_;
+    }
+
+private:
+    std::uint64_t value_ = 0;
+    std::mutex mutex_;
+};
```
rwalworth, line 226: **todo:** `NonceStore` uses `std::uint64_t` but the includes only list `<mutex>` and `<string>`. `<cstdint>` is missing.

```cpp
#include <cstdint>  // add this
#include <mutex>
#include <string>
```

guides/api-best-practices-cpp.md
```diff
@@ -1,24 +1,287 @@
...
+| `map<KEY, VALUE>` | `std::map<KEY, VALUE>`                        | `std::map<KEY, VALUE>`                            | Use `std::unordered_map` if needed      |
+| `type`            | `std::type_index`                             | `std::type_index`                                 | From `<typeindex>`                      |
+| `uuid`            | UUID library type                             | UUID library type                                 | No standard UUID until C++26            |
+| `date`            | Custom date struct or library type            | `std::chrono::year_month_day`                     | `year_month_day` is C++20               |
+| `time`            | Custom time struct or library type            | `std::chrono::hh_mm_ss<std::chrono::nanoseconds>` | `hh_mm_ss` is C++20                     |
+| `dateTime`        | `std::chrono::system_clock::time_point`       | `std::chrono::system_clock::time_point`           | -                                       |
+| `zonedDateTime`   | `time_point` plus explicit IANA zone string   | `std::chrono::zoned_time`                         | `zoned_time` is C++20                   |
```
rwalworth: **suggestion (non-blocking):** The C++17 entries for `date` ("Custom date struct or library type") and `zonedDateTime` ("time_point plus explicit IANA zone string") are vague. The current SDK doesn't use calendar date types at all - it only uses `std::chrono::system_clock::time_point` and `std::chrono::system_clock::duration` - so this may not come up in practice.

But if V3 does need a calendar date or timezone-aware type, the answer is the **Howard Hinnant date library** (`github.com/HowardHinnant/date`). It's the de facto C++17 standard: `std::chrono::year_month_day` and `std::chrono::zoned_time` in C++20 _are_ this library, slightly revised. The C++17 equivalents are `date::year_month_day` and `date::zoned_time<Duration>`.

"Custom date struct or library type" could reasonably name that library. "time_point plus explicit IANA zone string" isn't a type (it's a convention) and should be replaced with `date::zoned_time<Duration>` from the same library.

rwalworth (changes requested):
Thanks for the detailed work here, @Omswastik-11. The structure is solid and most sections are accurate. Honestly, reviewing this spurred me to re-examine current C++17 best practices and in doing so I realized there are a few places where the V2 SDK deviates from what's considered industry standard - so some of my comments are as much about updating the guide to reflect the right V3 direction as they are about the content you wrote.

The purpose of this guide is to document current C++17 best practice, not to describe what V2 does. Where V2 takes a shortcut, the guide should note it and point toward the better approach. A few sections currently document the V2 pattern without flagging that it's not the recommended way forward. I've left detailed comments below.

Omswastik-11:
Thanks @rwalworth for the review . I will implement them .

rwalworth (approved):
Thanks for working through all the feedback, @Omswastik-11 - the result type model, executor async pattern, synchronization examples, and include fixes all look right. Merging!

PR #192, opened 2026-04-14
ci(spec-lint): add scoped DSL lint script and workflow

### LINKED ISSUE #189: [V3 Docs] Spec-lint CI: lint meta-language DSL blocks across markdown while ignoring language-specific examples
## Summary

I propose adding a lightweight spec-lint check that prevents known meta-language regressions (e.g., `byte[]`, `boolean`, `Set<>`, `long`, hyphenated namespace identifiers) while remaining safe for docs that include code examples in many languages (Python, Rust, Java, JS, Go, etc.).

The key design point is scope:
- detect markdown changes broadly
- lint only fenced code blocks that look like the API meta-language DSL
- skip regular programming-language example blocks

## Why this is needed

We have seen recurring spec regressions in prototype/API DSL content:
- Java-specific types (`byte[]`, `boolean`, `String`, `Set<>`)
- undefined primitive (`long`) in DSL
- typo patterns like `RawFormate`
- namespace identifier hygiene issues

A lightweight CI guard would catch these early and reduce review churn.

## Proposed behavior

1. CI runs when markdown changes.
2. Script scans fenced code blocks.
3. Script lints only likely DSL blocks (e.g., containing `namespace`, `requires`, `abstraction`, `enum`, `constant`, `extends`, `@@...`).
4. Script ignores known language-tagged example blocks (`python`, `java`, `rust`, `go`, `js`, etc.) to avoid false positives.
5. Existing rule set remains regex-based and intentionally lightweight.

## Initial rule set

- forbid `byte[]` -> use `bytes`
- forbid `boolean` -> use `bool`
- forbid `String` -> use `string`
- forbid `Set<...>` -> use `set<TYPE>`
- forbid `long` in DSL fields/returns -> use explicit `intX` (e.g., `int64`)
- forbid hyphen in namespace identifiers
- catch typo `RawFormate`

## Open questions for maintainers

1. Should CI trigger for all markdown changes, or only selected directories?
2. Should we keep this as a warning-only discussion first, then enforce failing CI?
3. Any additional forbidden patterns to include in phase 1?
4. Should we maintain a small allowlist/ignore file for edge cases?

## Acceptance criteria

- Maintainers agree on lint scope and enforcement level
- Script safely ignores non-DSL code examples in language-specific docs
- CI catches known DSL regression patterns with low false positives
- Follow-up PR introduces script + workflow with agreed scope

Omswastik-11:
Closes #189

### DIFF
--- a/.github/workflows/spec-lint.yml
+++ b/.github/workflows/spec-lint.yml
@@ -52,8 +52,10 @@ jobs:
 
       - name: Run spec lint
         if: steps.changed.outputs.files != ''
+        env:
+          CHANGED_FILES: ${{ steps.changed.outputs.files }}
         run: |
-          python scripts/spec_lint.py ${{ steps.changed.outputs.files }}
+          python scripts/spec_lint.py $CHANGED_FILES
 
       - name: No relevant file changes
         if: steps.changed.outputs.files == ''
--- a/.github/workflows/spec-lint.yml
+++ b/.github/workflows/spec-lint.yml
@@ -1,0 +1,60 @@
+name: Spec Lint
+
+on:
+  pull_request:
+    paths:
+      - v3-sandbox/prototype-api/*.md
+      - proposals/hips/*.md
+      - scripts/spec_lint.py
+      - .spec-lint-ignore
+      - .github/workflows/spec-lint.yml
+  push:
+    branches:
+      - main
+    paths:
+      - v3-sandbox/prototype-api/*.md
+      - proposals/hips/*.md
+      - scripts/spec_lint.py
+      - .spec-lint-ignore
+      - .github/workflows/spec-lint.yml
+
+jobs:
+  spec-lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Determine changed scoped markdown files
+        id: changed
+        shell: bash
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            BASE_SHA="${{ github.event.pull_request.base.sha }}"
+          else
+            BASE_SHA="${{ github.event.before }}"
+          fi
+
+          if [[ -z "$BASE_SHA" || "$BASE_SHA" == "0000000000000000000000000000000000000000" ]]; then
+            BASE_SHA="$(git rev-list --max-parents=0 HEAD)"
+          fi
+
+          FILES="$(git diff --name-only "$BASE_SHA" "${{ github.sha }}" -- 'v3-sandbox/prototype-api/*.md' 'proposals/hips/*.md' | tr '\n' ' ')"
+          echo "files=$FILES" >> "$GITHUB_OUTPUT"
+
+      - name: Run spec lint
+        if: steps.changed.outputs.files != ''
+        run: |
+          python scripts/spec_lint.py ${{ steps.changed.outputs.files }}
+
+      - name: No relevant file changes
+        if: steps.changed.outputs.files == ''
+        run: echo "No scoped markdown files changed."
--- a/scripts/spec_lint.py
+++ b/scripts/spec_lint.py
@@ -1,0 +1,289 @@
+#!/usr/bin/env python3
+"""Lightweight lint checks for API meta-language markdown specs.
+
+This script scans markdown fenced code blocks and lints only blocks that look
+like the Hiero API meta-language DSL. This avoids false positives in regular
+language examples (Python, Java, Rust, Go, JavaScript, etc.).
+"""
+
+from __future__ import annotations
+
+import argparse
+import fnmatch
+import re
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Iterable, List, Sequence
+
+
+@dataclass(frozen=True)
+class Rule:
+    rule_id: str
+    pattern: re.Pattern[str]
+    message: str
+
+
+@dataclass(frozen=True)
+class Finding:
+    file_path: Path
+    line_number: int
+    rule_id: str
+    message: str
+    line_text: str
+
+
+RULES: List[Rule] = [
+    Rule(
+        rule_id="java-byte-array",
+        pattern=re.compile(r"\bbyte\[\]"),
+        message="Use meta-language type `bytes` instead of Java-specific `byte[]`.",
+    ),
+    Rule(
+        rule_id="java-boolean",
+        pattern=re.compile(r"\bboolean\b"),
+        message="Use canonical meta-language type `bool` instead of `boolean`.",
+    ),
+    Rule(
+        rule_id="java-string",
+        pattern=re.compile(r"\bString\b"),
+        message="Use canonical meta-language type `string` instead of `String`.",
+    ),
+    Rule(
+        rule_id="java-set-generic",
+        pattern=re.compile(r"\bSet<"),
+        message="Use canonical collection type `set<TYPE>` instead of `Set<TYPE>`.",
+    ),
+    Rule(
+        rule_id="undefined-long-field",
+        pattern=re.compile(r":\s*long\b"),
+        message="Use an explicit meta-language integer type like `int64` instead of `long`.",
+    ),
+    Rule(
+        rule_id="undefined-long-return",
+        pattern=re.compile(r"^\s*long\b"),
+        message="Use an explicit meta-language integer type like `int64` instead of `long`.",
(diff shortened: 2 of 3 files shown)


Omswastik-11:
Hi @rwalworth . I have implemented the draft .Let me know if you want any changes or any additions in Rules .

.github/workflows/spec-lint.yml
```diff
@@ -0,0 +1,60 @@
...
+            BASE_SHA="$(git rev-list --max-parents=0 HEAD)"
+          fi
+
+          FILES="$(git diff --name-only "$BASE_SHA" "${{ github.sha }}" -- 'v3-sandbox/prototype-api/*.md' 'proposals/hips/*.md' | tr '\n' ' ')"
+          echo "files=$FILES" >> "$GITHUB_OUTPUT"
+
+      - name: Run spec lint
+        if: steps.changed.outputs.files != ''
+        run: |
+          python scripts/spec_lint.py ${{ steps.changed.outputs.files }}
```
rwalworth: **issue (blocking, security):** `${{ steps.changed.outputs.files }}` is used directly inside a `run:` script, which is a GitHub Actions script injection anti-pattern. GitHub Actions expressions in `run:` are substituted verbatim before the shell executes - if the value ever contained shell metacharacters, they would be interpreted.

The fix is to pass the value through an `env:` variable instead:

```yaml
- name: Run spec lint
  if: steps.changed.outputs.files != ''
  env:
    CHANGED_FILES: ${{ steps.changed.outputs.files }}
  run: |
    python scripts/spec_lint.py $CHANGED_FILES
```

The actual injection risk here is low (filenames come from a filtered `git diff --name-only` on controlled paths), but the pattern should be avoided - it's flagged by tools like `zizmor` and goes against GitHub's own security guidance for Actions workflows.

rwalworth (changes requested):
Thanks for putting this together, @Omswastik-11 - this is exactly what I had in mind from the discussion on #189. The script is well-structured, the rule set matches what we agreed on, and the scoped detection logic handles the false-positive problem cleanly.

One thing to fix before I can merge - left a comment on the workflow file below. It's a small change. Once that's addressed, we're good to go!

I'm also going to ask @exploreriii to review the Python code.

rwalworth (approved):
LGTM! The security fix looks exactly right - passing `steps.changed.outputs.files` through `CHANGED_FILES` in the `env:` block is the correct mitigation. 

I'm still going to ask @exploreriii to review the Python code before we get this merged.

.github/workflows/spec-lint.yml
```diff
@@ -0,0 +1,62 @@
...
+      - .spec-lint-ignore
+      - .github/workflows/spec-lint.yml
+
+jobs:
+  spec-lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
exploreriii, line 27: Could you please read
https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/guides/issue-progression/for-developers/github-action-workflows.md#how-to-pin-github-actions-to-a-specific-commit-hash
We should always pin, also your packets are AI generated and out dated, please use the latest version by researching the packet github page and seeing the latest instructions
https://github.com/actions/checkout

.github/workflows/spec-lint.yml
```diff
@@ -0,0 +1,62 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
```
exploreriii, line 34: out dated/etc

.github/workflows/spec-lint.yml
```diff
@@ -0,0 +1,62 @@
...
+      - v3-sandbox/prototype-api/*.md
+      - proposals/hips/*.md
+      - scripts/spec_lint.py
+      - .spec-lint-ignore
+      - .github/workflows/spec-lint.yml
+
+jobs:
+  spec-lint:
+    runs-on: ubuntu-latest
+
```
exploreriii, line 24: you should harden the runner, please study examples eg at https://github.com/hiero-ledger/hiero-sdk-python/tree/main/.github/workflows

.github/workflows/spec-lint.yml
```diff
@@ -0,0 +1,62 @@
...
+    branches:
+      - main
+    paths:
+      - v3-sandbox/prototype-api/*.md
+      - proposals/hips/*.md
+      - scripts/spec_lint.py
+      - .spec-lint-ignore
+      - .github/workflows/spec-lint.yml
+
+jobs:
```
exploreriii, line 21: you should restrict the permissions of a workflow, again, study examples or read
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax

.github/workflows/spec-lint.yml
```diff
@@ -0,0 +1,62 @@
...
+  spec-lint:
+    runs-on: ubuntu-latest
+
+    steps:
```
exploreriii, line 25: should add concurrency as a guard

scripts/spec_lint.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python3
+"""Lightweight lint checks for API meta-language markdown specs.
```
exploreriii, line 2: I would suggest making this a file, and inside that different chunks eg
scripts/spec_lint/
  __main__.py
  rules.py
  markdown_blocks.py
  config.py
as i assume this will get bigger

or
  rules/
    java.py
    javascript.py
etc

scripts/spec_lint.py
```diff
@@ -0,0 +1,300 @@
...
+    ),
+    Rule(
+        rule_id="java-boolean",
+        pattern=re.compile(r"\bboolean\b"),
+        message="Use canonical meta-language type `bool` instead of `boolean`.",
+    ),
+    Rule(
+        rule_id="java-string",
+        pattern=re.compile(r"\bString\b"),
+        message="Use canonical meta-language type `string` instead of `String`.",
```
exploreriii, line 50: if this is meta language, should this be named java-x?

scripts/spec_lint.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python3
+"""Lightweight lint checks for API meta-language markdown specs.
+
+This script scans markdown fenced code blocks and lints only blocks that look
+like the Hiero API meta-language DSL. This avoids false positives in regular
```
exploreriii, line 5: This docstring should be longer in my opinion, is quite complicated what you do and also there are some simplifications to note

scripts/spec_lint.py
```diff
@@ -0,0 +1,300 @@
...
+    re.compile(r"\bnamespace\b"),
+    re.compile(r"\brequires\b"),
+    re.compile(r"\babstraction\b"),
+    re.compile(r"\benum\b"),
+    re.compile(r"\bconstant\b"),
+    re.compile(r"\bextends\b"),
+    re.compile(r"@@[A-Za-z]"),
+)
+
+DEFAULT_SCOPE_GLOBS: Sequence[str] = (
```
exploreriii, line 123: must you determine the changed files in the workflow since you have this?

scripts/spec_lint.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python3
+"""Lightweight lint checks for API meta-language markdown specs.
+
+This script scans markdown fenced code blocks and lints only blocks that look
+like the Hiero API meta-language DSL. This avoids false positives in regular
+language examples (Python, Java, Rust, Go, JavaScript, etc.).
+"""
```
exploreriii, line 7: Right now you are guessing, perhaps require and check an explicit fence language for DSL blocks
e.g.
DSL_FENCE_LANGS = {"hiero", "hiero-dsl", "api-spec"}

def should_lint_block(fence_lang: str, block_lines: Sequence[str], untagged_is_dsl: bool) -> bool:
    """Return True only for explicitly tagged API meta-language DSL blocks."""


Only fenced code blocks tagged `hiero`, `hiero-dsl`, or `api-spec` are linted.
Untagged code blocks are ignored.

what do you think?

exploreriii (changes requested):
i think this will be quite difficult to make accurate - i can imagine some false positives and false negatives.
One idea is to split this into the most easily maintainable parts, with good documentation for each part. Should aim for the minimum amount that is accurate, better to merge and improve on it slowly that then create risk

Some issues to think about
-broad regex rules that match in strings, comments, or prose
- can incorrectly lint non-DSL examples
- any block tagged java, python etc is ignored

Also left some feedback on workflow structure and naming which is not following best practices

exploreriii (commented):
if you go ahead with this, I imagine you'll need really robust tests that the linter picks up the DSL and does not do unexpected things

PR #194, opened 2026-04-16
docs: enhance API guidelines with new type definitions and service API

hendrikebbers:
service layer / enterprise layer spec started


guides/api-guideline.md
```diff
@@ -202,6 +203,45 @@ abstraction FruitFactory<$$Product extends Fruit> {
...
+}
+```
+
+In this example, `constructorParams` accepts a heterogeneous list of `ContractParam` instances — each element may be
+a different concrete instantiation (e.g., `ContractParam<int32>`, `ContractParam<string>`).
+
+Rules:
+
+- `ANY` can only appear as a type argument, never as a standalone type (e.g., a parameter type, return type, or field
+  type cannot be just `ANY`).
```
rwalworth, line 233: **issue:** The first rule here contradicts the paragraph that immediately follows it (and the basic data types table entry on line 51).

The rule currently says:

> `ANY` can only appear as a type argument, never as a standalone type (e.g., a parameter type, return type, or field type cannot be just `ANY`).

But the very next paragraph says:

> `ANY` can also appear as a standalone type (e.g., a parameter or return type), in which case it acts as the language's top type.

The basic data types table also lists `ANY` as a valid standalone type with the note "Use sparingly (see best practices)".

The intent seems to be that standalone `ANY` is *discouraged* but not *forbidden* - so the rule bullet should be softened. Something like:

> Prefer `ANY` as a wildcard type argument. Using `ANY` as a standalone type (e.g., a parameter type or return type) is strongly discouraged — see [Avoid `ANY` as a standalone type](#avoid-any-as-a-standalone-type).

Then the closing paragraph below the rules becomes redundant and can be removed.

v3-sandbox/prototype-api/service.md
```diff
@@ -0,0 +1,49 @@
...
+```
+namespace enterprise.service.contract
+requires common
+
+ParamSupplier<$$Type> {
+    //TODO: add a generic param for it? In java a sealed class would be nice. We need to investigate more
+    ANY getNativeContractValue()
+}
+
+Param<$$Type>(
```
rwalworth:
```suggestion
Param<$$Type> {
```

v3-sandbox/prototype-api/service.md
```diff
@@ -0,0 +1,49 @@
...
+
+ContractCallResult {
+    //TODO: Provide a good and extentsible way to receive 0-N of the possible return types
+}
+
+SmartContractService {
+
+    @@throws(service-error) common.ContractId createContract(fileId:common.FileId, constructorParams:Param<ANY>...)
+    
+    @@throws(service-error) common.ContractId createContract(contents:bytes, , constructorParams:Param<ANY>...)
```
rwalworth:
```suggestion
    @@throws(service-error) common.ContractId createContract(contents:bytes, constructorParams:Param<ANY>...)
```

v3-sandbox/prototype-api/service.md
```diff
@@ -0,0 +1,49 @@
...
+Param<string> ofBytes23(value:string)
+Param<bytes> ofBytes(value:bytes)
+Param<bytes> ofBytes23(value:bytes)
+Param<string> ofAddress(value:string)
+Param<common.Address> ofAddress(value:common.Address)
+Param<boolean> ofBool(value:boolean)
+Param<uint8> uint8(value:uint8)
+Param<int8> int8(value:int8)
+//TODO
+Param<uint256> uint8(value:uint256)
```
rwalworth:
```suggestion
Param<uint256> uint256(value:uint256)
```

v3-sandbox/prototype-api/service.md
```diff
@@ -0,0 +1,49 @@
...
+Param<int256> int8(value:int256)
```
rwalworth:
```suggestion
Param<int256> int256(value:int256)
```

v3-sandbox/prototype-api/service-contract.md
rwalworth, line 1: This file should be `service-contract.md`

v3-sandbox/prototype-api/service.md
```diff
@@ -0,0 +1,49 @@
...
+ParamSupplier<$$Type> {
+    //TODO: add a generic param for it? In java a sealed class would be nice. We need to investigate more
+    ANY getNativeContractValue()
+}
+
+Param<$$Type>(
+    @@immutable value:$$Type
+    @@immutable nativeType:string
+    @@immutable supplier:ParamSupplier<$$Type>
+}
```
rwalworth: **thought:** I'm wondering whether `abstraction ContractParam` would be a better fit for the smart contract service for now, and we hold off on `ANY` until we have use cases where an abstraction simply can't express what we need - something like:

```
abstraction ParamSupplier<$$Type> {
    $$Type getNativeContractValue()    // $$Type instead of ANY — no wildcard needed
}

abstraction ContractParam {
    @@immutable nativeType: string
}

TypedParam<$$Type> extends ContractParam {
    @@immutable value: $$Type
    @@immutable supplier: ParamSupplier<$$Type>
}

// method signatures become:
createContract(fileId: common.FileId, constructorParams: ContractParam...)
createContract(contents: bytes, constructorParams: ContractParam...)
callContractFunction(contractId: common.ContractId, functionName: string, params: ContractParam...)
```

The factory methods (`ofString`, `ofUint256`, etc.) would return `TypedParam<string>`, `TypedParam<uint256>`, etc., all of which are `ContractParam` so callers use them identically. The advantage is that `ContractParam...` maps cleanly to an interface/trait/protocol in every language, including future ones. Wildcard generics (`Param<ANY>`, `Param<?>`) don't exist in all our target languages and would require language-specific mapping guidance for Go, Rust, C++, and anything that comes after.

rwalworth (dismissed):

rwalworth (approved):

PR #195, opened 2026-04-16
docs: add HIP-1137 response codes to SDK design doc

rwalworth:
## Summary

Updates the HIP-1137 SDK design doc to document the seven new response codes introduced by the consensus node implementation of registered node transactions. The previous text incorrectly stated that HIP-1137 defines no new response codes.

### Response Codes

Replaced the placeholder note in the Internal Changes / Response Codes section with the full set of registered-node-specific response codes (529–535) and a transaction retry note.

| Code | Value | Description |
|------|-------|-------------|
| `INVALID_REGISTERED_NODE_ID` | 529 | Registered node ID is invalid or does not exist |
| `INVALID_REGISTERED_ENDPOINT` | 530 | Port is out of range or address field is not set |
| `REGISTERED_ENDPOINTS_EXCEEDED_LIMIT` | 531 | Service endpoint list exceeds the 50-entry limit |
| `INVALID_REGISTERED_ENDPOINT_ADDRESS` | 532 | IP address is not 4/16 bytes, or domain name is not a valid ASCII FQDN |
| `INVALID_REGISTERED_ENDPOINT_TYPE` | 533 | Endpoint does not specify an endpoint type |
| `REGISTERED_NODE_STILL_ASSOCIATED` | 534 | Node cannot be deleted; still referenced by a consensus node |
| `MAX_REGISTERED_NODES_EXCEEDED` | 535 | `associatedRegisteredNodes` list exceeds the 20-entry limit |

## Files Changed Summary

| File | Change |
|------|--------|
| `proposals/hips/hip-1137.md` | Response Codes section replaced with full code list and retry guidance |

## Breaking Changes

**None.** Documentation-only change.

### DIFF
--- a/proposals/hips/hip-1137.md
+++ b/proposals/hips/hip-1137.md
@@ -227,7 +227,17 @@ All three registered node transactions (`RegisteredNodeCreateTransaction`, `Regi
 
 ### Response Codes
 
-HIP-1137 does not define any new response codes. If response codes specific to registered node transactions are introduced during consensus node implementation, this document should be updated to document them and evaluate any transaction retry implications for the SDKs.
+- `INVALID_REGISTERED_NODE_ID` — The registered node ID is invalid or does not exist in network state.
+- `INVALID_REGISTERED_ENDPOINT` — A service endpoint is invalid; the port is out of range, or no address field is set.
+- `REGISTERED_ENDPOINTS_EXCEEDED_LIMIT` — The service endpoint list exceeds the maximum of 50 entries.
+- `INVALID_REGISTERED_ENDPOINT_ADDRESS` — A service endpoint has an invalid address; the IP address is not 4 (IPv4) or 16 (IPv6) bytes, or the domain name is not a valid ASCII FQDN.
+- `INVALID_REGISTERED_ENDPOINT_TYPE` — A service endpoint does not specify an endpoint type.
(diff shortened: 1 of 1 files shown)


ivaylonikolov7 (approved):

ivaylonikolov7:
I don't know about the other SDKs but don't you have a script that automatically adds them to your SDKs when they the response codes are updated? Do yo we need to have them specifically added in a design doc?

Dosik13:
In the Go sdk there is no such script, but even if there was, I believe it is nice to have the response codes in the design docs.

rwalworth:
> I don't know about the other SDKs but don't you have a script that automatically adds them to your SDKs when they the response codes are updated? Do yo we need to have them specifically added in a design doc?

Yeah I was thinking the individual codes don't necessarily need to be enumerated in the doc - I was thinking about changing the design doc template to remove these. However, what does need to be here is the retry guidance, since that behavioral decision can't be inferred from the proto definitions alone (none of these codes represent transient failures, so SDKs should not retry on any of them).

I do think the test cases need the specific response code expected in each error scenario. The PR also updates several test cases to call out the exact expected status (e.g. "fails with a receipt status of `INVALID_REGISTERED_NODE_ID`") - that specificity is important so SDK implementors know exactly what to assert against.

PR #254, opened 2026-05-08
fix: mirror node ingress endpoint standartization

venilinvasilev:
### Description

Adds a design proposal documenting how the Hiero SDKs should stop rewriting Mirror Node URL ports for local development and CI. Today, some SDKs mutate the configured Mirror Node URL at multiple distinct call sites when the host is `localhost` / `127.0.0.1` — for example, rewriting port `5551` to `8084` for Java REST API calls (fee estimation, registered-node address book queries) and rewriting `5551` to `8545` for Mirror Node Web3 contract queries. The `8084` set is actively growing as the Mirror Node team migrates more endpoints from the legacy JS REST API service to the Java API service. The proposal removes this coupling to Mirror Node's internal service topology by relying on the single Solo ingress endpoint that fronts every Mirror Node service.

### Covers

- An end-state rule that the user-supplied Mirror Node URL must be used as-is — no SDK code path may mutate it based on `localhost` / `127.0.0.1` or any other characteristic of the configured value.
- Update of each SDK's local mirror network preset to point at the Solo ingress endpoint.
- Hard prerequisite that local and CI Solo networks (`hiero-solo-action`) must be started with the ingress controller enabled.
- A note that the rollout is gated on [`hiero-solo-action#120`](https://github.com/hiero-ledger/hiero-solo-action/issues/120) and the per-SDK atomic PR shape for once the gate lifts.

### DIFF
--- a/proposals/mirror-node-ingress-endpoint-standardization.md
+++ b/proposals/mirror-node-ingress-endpoint-standardization.md
@@ -11,27 +11,23 @@ Mirror Node URL from the user, and that single URL is what should be used everyw
 rewriting.
 
 The Mirror Node team is migrating endpoints from the legacy JS REST API service to the new Java API service. As a
-result, individual Mirror Node operations live behind different internal ports, and the SDKs have grown ad-hoc
-port-rewrite logic to reach the right service. The JavaScript SDK is the most visible offender. For example, when
-the configured `mirrorNodeUrl` points at `localhost`, the SDK currently does things like:
-
-```js
-// Current behavior in some SDK paths:
-if (mirrorNodeUrl.includes("localhost") || mirrorNodeUrl.includes("127.0.0.1")) {
-    mirrorNodeUrl = mirrorNodeUrl.replace("5551", "8384");
-}
-```
+result, individual Mirror Node operations live behind different internal ports, and some SDKs have grown ad-hoc
+port-rewrite logic to reach the right service. These rewrites sit at multiple distinct call sites — not in a single
+block — and each call site targets a different Mirror Node service. For example, requests served by the Java API
+(such as fee estimation and registered-node address book queries) rewrite the configured port `5551` to `8084`,
+while Mirror Node Web3 contract queries rewrite `5551` to `8545`. The set of Java-API-bound rewrites is actively
+growing as the Mirror Node team migrates more endpoints away from the legacy JS REST API service.
 
 This pattern is fragile. It hardcodes Mirror Node's internal service topology into the SDKs; every time Mirror Node
-consolidates, renames, or moves a service port, every SDK has to chase the change. The rewrite only fires for
+consolidates, renames, or moves a service port, every SDK has to chase the change. The rewrites only fire for
 `localhost` / `127.0.0.1`, which means production-like setups using a real ingress already work without rewriting —
 confirming that rewriting is purely a workaround for the local/CI case.
 
 [Solo](https://github.com/hiero-ledger/hiero-solo-action) exposes Mirror Node behind a **single ingress endpoint**
 that fronts every Mirror Node service on a single port. The SDKs will treat the user-supplied Mirror Node URL as
-authoritative, remove all in-SDK port-rewrite logic, and update the `MirrorNetwork.LOCAL_NODE` constants (and
-language equivalents) to point at the Solo ingress port. CI does not need separate work: each SDK's CI pipeline
-spins up the same local network the developer runs locally, so fixing the SDK source code fixes both.
+authoritative, remove all in-SDK port-rewrite logic, and update each SDK's local mirror network preset to point at
+the Solo ingress endpoint. CI does not need separate work: each SDK's CI pipeline spins up the same local network
+the developer runs locally, so fixing the SDK source code fixes both.
 
 **Once this proposal is implemented, every Solo network used for SDK development must be started with the ingress
 controller enabled.** This applies in both directions:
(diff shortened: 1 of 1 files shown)


proposals/mirror-node-ingress-endpoint-standardization.md
```diff
@@ -0,0 +1,129 @@
...
+**Once this proposal is implemented, every Solo network used for SDK development must be started with the ingress
+controller enabled.** This applies in both directions:
+
+- **Local development** — when running Solo on a developer machine, the network must be started with the ingress
+  controller enabled.
+- **CI** — when invoking [`hiero-solo-action`](https://github.com/hiero-ledger/hiero-solo-action), the workflow must
+  enable the ingress controller.
+
+Without the ingress controller enabled, the Mirror Node services that the SDK now reaches via the ingress port will
+not be exposed and integration tests will fail.
```
rwalworth, line 41: **suggestion (non-blocking)**: This section establishes the "ingress controller must be enabled" prerequisite but doesn't tell each SDK owner how to land the cleanup safely, and doesn't make the prerequisite's actual status explicit. Worth being concrete about both:

The prerequisite is real and unfinished. `hiero-solo-action#120` is still open with unresolved design questions ("Which local port should ingress be forwarded to?" / "Should ingress be enabled by default or behind a flag?"). The latest action version (v0.19.0) only enables the ingress controller as a side effect of `installRelay: true`; there's no first-class input for it yet. Until the PR lands and a corresponding action release ships, no SDK can complete the cleanup this proposal describes. The doc should say that explicitly so it's clear the rollout is gated.

Once that prerequisite is met, the cleanup can be a clean per-SDK PR rather than a cross-repo coordinated rollout - every SDK's CI already goes through `hiero-solo-action`, so a single atomic PR per SDK can (a) bump or reconfigure `hiero-solo-action` to enable the ingress controller, (b) remove the in-SDK port-rewrite logic, and (c) update the SDK's local mirror network preset to point at the ingress endpoint. Because each SDK's CI uses the same Solo network developers run locally, this keeps source and CI in sync without further coordination.

It might be a good idea to add a couple sentences (or a subsection) about the rollout. The doc should make the gating dependency obvious and gives SDK owners a clean checklist for when the gate lifts.

proposals/mirror-node-ingress-endpoint-standardization.md
```diff
@@ -0,0 +1,129 @@
...
+```js
+// REMOVE: ad-hoc port rewriting that hardcodes Mirror Node service topology.
+if (mirrorNodeUrl.includes("localhost") || mirrorNodeUrl.includes("127.0.0.1")) {
+    mirrorNodeUrl = mirrorNodeUrl.replace("5551", "8384");
+}
+```
+
+Each SDK owner is responsible for auditing their codebase for equivalent rewrites — including any path that swaps
+between gRPC and REST ports for the same Mirror Node URL — and removing them. After this change, the configured URL
+is the single source of truth for every Mirror Node service the SDK calls.
```
rwalworth: **suggestion (non-blocking):** The prescription "every code path... must be deleted" implicitly assumes every SDK has a `localhost`-guarded `replace()` to delete. Some SDKs (e.g., ones that already do the right thing, or ones that compute Mirror Node endpoints via a per-network map rather than mutation) won't fit that exact shape, which makes "delete this block" awkward.

Suggest reframing the requirement abstractly as an end-state rule:

> The user-supplied Mirror Node URL must be used as-is for every Mirror Node operation. No SDK code path may mutate the URL (host, port, scheme, or path prefix) based on whether the host is `localhost` / `127.0.0.1` or any other characteristic of the configured value.

Then "audit your SDK and remove anything that violates this" naturally covers the `if (localhost) replace(...)` shape, the per-network port-translation map shape, and the "we already do this correctly, no change needed" case - without enumerating SDK specifics.

proposals/mirror-node-ingress-endpoint-standardization.md
```diff
@@ -0,0 +1,129 @@
...
+
+This design proposal eliminates **in-SDK rewrites of the Mirror Node URL port** for queries that target services
+hosted on different Mirror Node ports in local development and Continuous Integration (CI) environments. It does not
+introduce a new endpoint, a new transport, or any new public configuration: the SDK already accepts a single
+Mirror Node URL from the user, and that single URL is what should be used everywhere — without internal port
+rewriting.
+
+The Mirror Node team is migrating endpoints from the legacy JS REST API service to the new Java API service. As a
+result, individual Mirror Node operations live behind different internal ports, and the SDKs have grown ad-hoc
+port-rewrite logic to reach the right service. The JavaScript SDK is the most visible offender. For example, when
```
rwalworth: **nitpick:** Small phrasing - calling JS "the most visible offender" anchors the doc on one SDK. Could just say "some SDKs" and let the JS code block immediately after carry the specificity.

proposals/mirror-node-ingress-endpoint-standardization.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+This pattern is fragile. It hardcodes Mirror Node's internal service topology into the SDKs; every time Mirror Node
+consolidates, renames, or moves a service port, every SDK has to chase the change. The rewrite only fires for
+`localhost` / `127.0.0.1`, which means production-like setups using a real ingress already work without rewriting —
+confirming that rewriting is purely a workaround for the local/CI case.
+
+[Solo](https://github.com/hiero-ledger/hiero-solo-action) exposes Mirror Node behind a **single ingress endpoint**
+that fronts every Mirror Node service on a single port. The SDKs will treat the user-supplied Mirror Node URL as
+authoritative, remove all in-SDK port-rewrite logic, and update the `MirrorNetwork.LOCAL_NODE` constants (and
```
rwalworth: **nitpick:** The prescription here names a JS constant (`MirrorNetwork.LOCAL_NODE`). Reframing as "the SDK's local mirror network preset" would keep the prescription itself generic - the JS specifics in the per-SDK section already carry the example.

rwalworth (commented):
I like this proposal. The end state of "stop rewriting Mirror Node ports inside the SDK; rely on a single Solo ingress endpoint" is the right call. Left a few non-blocking suggestions inline, mostly around making the framing language-agnostic and tightening the rollout story.

One thing I wanted to flag: the proposal cites `mirrorNodeUrl.replace("5551", "8384")` as the example pattern, but I couldn't find `8384` anywhere - actual code in JS / Go / Swift uses `8084` (the Java REST API port). Typo, or am I missing a port? If it's a typo it shows up in the PR body and twice in the proposal body, so worth fixing in all three.

proposals/mirror-node-ingress-endpoint-standardization.md
venilinvasilev: I've now made it clear in the internal changes section.

proposals/mirror-node-ingress-endpoint-standardization.md
venilinvasilev: This has been addressed now.

proposals/mirror-node-ingress-endpoint-standardization.md
venilinvasilev: done

proposals/mirror-node-ingress-endpoint-standardization.md
venilinvasilev: done

ddeskov-limechain (approved):
LGTM

rwalworth (approved):
LGTM, thanks @venilinvasilev!

PR #257, opened 2026-05-11
docs: propose opt-in gas estimation for ContractCallQuery

ddeskov-limechain:
## Summary

Adds a design proposal at `proposals/contract-call-query-gas-estimation.md`
fixing `ContractCallQuery.execute(client)` when the caller does not call
`setGas(...)`. Today the SDK's internal `getCost()` sends a `COST_ANSWER`
request with `gas=0` and the consensus node correctly rejects with
`INSUFFICIENT_GAS` — the failure is from the SDK's hidden cost probe, not
the user's call.

Upstream issues:
- hiero-ledger/hiero-sdk-js#2848
- hiero-ledger/hiero-consensus-node#4854

## Design

Cross-SDK, opt-in, explicit-by-default. Three new entry points on
`ContractCallQuery`:

- `populateGasFromMirrorNode(client)` — async helper that fills gas via
  `MirrorNodeContractEstimateQuery`.
- `setAutomaticGasEstimation(enabled)` / `isAutomaticGasEstimation()` —
  opt-in flag; when `true`, `execute()` and `getCost()` internally call
  `populateGasFromMirrorNode` if gas is null.
- Behavior change: when gas is null and auto-estimation is off, `execute()`
  / `getCost()` throw `gas-not-set-error` *before* any network call, with a
  message pointing at the three remediation paths.

Sync setters stay sync. Async work lives in explicitly-async methods. No
hidden Mirror Node dependency for callers who already set gas. Scope is
`ContractCallQuery` only — other queries' cost flakiness is out of scope.

## Why not the alternatives

- **Auto-estimate by default** (the issue author's proposal): introduces a
  mandatory Mirror Node round-trip on every contract call query — hidden
  dependency, surprising failure modes when the mirror is unavailable.
- **`setCostAutomatic(true/false)` only** (the team's first idea): right
  intuition (opt-in), but the user actually needs to control *gas
  estimation*, not "cost"; with gas in hand the existing `getCost()`
  already works.

## Contents

- Summary, New APIs, Updated APIs, Internal Changes (Response Codes +
  Transaction Retry), Test Plan (8 G/W/T tests + TCK section), SDK Example
  — matches `templates/sdk-design-doc-template.md`.
- 3 Mermaid diagrams: current broken flow (sequence), proposed decision
  flow (flowchart), and the three caller-facing modes.
- All API definitions use the meta-language from `guides/api-guideline.md`.

## Follow-up

- Open a tracking issue in `hiero-sdk-tck` for the test plan once this
  proposal is approved.
- Per-SDK implementation issues to be filed against each SDK repo after
  approval.

### DIFF
--- a/proposals/contract-call-query-gas-estimation.md
+++ b/proposals/contract-call-query-gas-estimation.md
@@ -30,23 +30,27 @@ sequenceDiagram
     SDK-->>User: throws (looks like the user's call failed)
 ```
 
-This proposal fixes the bug across all Hiero SDKs by adopting **opt-in
-automatic gas estimation with explicit-by-default behavior**. Three new
-entry points are added to `ContractCallQuery`:
+This proposal fixes the bug across all Hiero SDKs with three changes:
 
-1. `populateGasFromMirrorNode(client)` — async helper that fills in gas via
-   `MirrorNodeContractEstimateQuery`.
-2. `setAutomaticGasEstimation(enabled)` / `isAutomaticGasEstimation()` —
-   opt-in flag; when `true`, `execute()` and `getCost()` internally call
-   `populateGasFromMirrorNode` if gas is null.
-3. A behavior change in `execute()` and `getCost()`: when gas is null and
-   automatic estimation is disabled, fail fast with a clear `gas-not-set`
-   error instead of dispatching a guaranteed-to-fail `COST_ANSWER` request.
+1. `ContractCallQuery.populateGasFromMirrorNode(client)` — async helper that
+   fills in gas via `MirrorNodeContractEstimateQuery` and stores it on the
+   query.
+2. `Client.setAutoEstimateContractCallGas(enabled)` /
+   `Client.isAutoEstimateContractCallGas()` — a **client-level** opt-in
+   flag. When `true`, `ContractCallQuery.execute()` and `getCost()` call
+   `populateGasFromMirrorNode(client)` internally if gas is null. The flag
+   lives on the client because auto-estimation is a property of how the
+   client handles contract reads, not of any one query — this matches how
+   relay-style auto-estimation works in practice.
+3. A behavior change in `ContractCallQuery.execute()` and `getCost()`:
+   when gas is null and the client flag is disabled, fail fast with a
+   clear `gas-not-set-error` instead of dispatching a guaranteed-to-fail
+   `COST_ANSWER` request.
 
 The design keeps the existing sync setters sync, isolates the async network
 round-trip in an explicitly-async method, and avoids introducing a hidden
-Mirror Node dependency on every contract call. Callers who already call
-`setGas(...)` see no change.
+Mirror Node dependency on every contract call unless the caller opts in at
+the client level. Callers who already call `setGas(...)` see no change.
 
 This proposal scopes the fix to `ContractCallQuery`. Cost-estimation
 flakiness reported for other queries (e.g., `AccountRecordsQuery`) has a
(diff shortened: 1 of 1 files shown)


proposals/contract-call-query-gas-estimation.md
```diff
@@ -0,0 +1,291 @@
...
+### `ContractCallQuery` — gas estimation helpers
+
+```
+ContractCallQuery {
+    // Returns the Mirror Node gas estimate for this query without mutating
+    // it. Uses MirrorNodeContractEstimateQuery against the client's mirror
+    // network. Does not contact the consensus node.
+    @@async
+    @@throws(mirror-node-error)
+    int64 estimateGasFromMirrorNode(client: Client)
```
rwalworth: **suggestion (blocking):** My preference would be to drop `estimateGasFromMirrorNode` and keep only `populateGasFromMirrorNode`.

To me, two methods that differ only in "returns the value" vs "stores the value" is API bloat for the sake of completeness. The "estimate, multiply by safety factor, then setGas" use case you'd be designing for seems rare - and `MirrorNodeContractEstimateQuery` is already public, so callers who genuinely want a raw estimate can use it directly (it's just one extra import). My read is: adding a wrapper for it on `ContractCallQuery` just because we can ends up multiplying every cross-SDK implementation by N for a use case most callers won't hit.

One obvious way to do it, in my view: if you don't want the SDK to apply the estimate, call `MirrorNodeContractEstimateQuery` yourself. If you do want it applied, call `populateGasFromMirrorNode(client)`. Done.

Happy to be talked out of this if there's a use case I'm missing.

proposals/contract-call-query-gas-estimation.md
```diff
@@ -0,0 +1,291 @@
...
+    // Convenience: runs estimateGasFromMirrorNode(client) and stores the
+    // result via setGas(...). Returns this for chaining.
+    @@async
+    @@throws(mirror-node-error)
+    ContractCallQuery populateGasFromMirrorNode(client: Client)
+
+    // Opt-in flag. Default: false. When true, execute() and getCost() call
+    // populateGasFromMirrorNode(client) internally if gas is null and
+    // queryPayment is not set.
+    ContractCallQuery setAutomaticGasEstimation(enabled: bool)
```
rwalworth: **issue (blocking):** My opinion is that `setAutomaticGasEstimation` doesn't belong on the query.

The flag is per-query, but to me the behaviour it controls - "implicitly hit the mirror node when the user didn't set gas" - feels like a property of how the client handles contract reads, not of any one query. Every call site has to make the same decision; my read is the flag is more likely to be set-and-forgotten than thoughtfully chosen per query.

Three options, ordered by my personal preference:

1. **Drop the flag.** Ship the bug fix + `populateGasFromMirrorNode`. Callers who want estimation chain one extra method; my view is this is what the EVM ecosystem expects anyway. Smallest API, no ambiguity.
2. **Move the flag to the client** (e.g. `client.setAutoEstimateContractCallGas(true)`). One toggle, applies to all `ContractCallQuery` executions on that client. To me this matches how relay-style auto-estimation actually lives at the client level.
3. **Keep it on the query** (current proposal). I'd want a strong argument for why per-query granularity matters in practice, which I don't see in the doc - but I'm open to it if you've got one.

If you keep option 3, my suggestion would be to have the rationale explicitly justify why this is per-query and not per-client. Without that, this reads to me as the weakest part of the design.

proposals/contract-call-query-gas-estimation.md
```diff
@@ -0,0 +1,291 @@
...
+   query returns the Mirror Node estimate and a subsequent `execute(client)`
+   (without re-populating) succeeds.
+5. **Given** a `ContractCallQuery` with `setAutomaticGasEstimation(true)`
+   and a Mirror Node that returns an error, **when** `execute(client)` runs,
+   **then** it surfaces `mirror-node-error` and never sends a `gas=0`
+   request to the consensus node.
+6. **Given** a `ContractCallQuery` with `setQueryPayment` set explicitly and
+   no `setGas`, **when** `execute(client)` runs, **then** it throws
+   `gas-not-set-error` (the queryPayment covers the SDK's fee for the query,
+   but the EVM call itself still needs gas).
```
rwalworth: **issue (blocking):** To my read, Test 6 contradicts the spec.

Test 6 says: `setQueryPayment` set + no `setGas` --> throws `gas-not-set-error`.

But the "Updated APIs" section says: "If gas is null and queryPayment is set: unchanged behavior (getCost is skipped, queryPayment is used directly)." If `getCost` is skipped, no `gas-not-set-error` gets thrown - my read is the SDK proceeds to the `ANSWER_ONLY` request with `gas=0`, and the consensus node returns `INSUFFICIENT_GAS` from the actual call (not the cost probe).

You'll need to pick one. My personal preference:

- **Tighten the spec.** Throw `gas-not-set-error` whenever gas is null, regardless of `queryPayment`. My view is that this matches the spirit of the proposal - "user didn't set gas" should never reach the wire, and letting it through under one branch undermines the rest of the design.

The other option - dropping Test 6 and treating queryPayment-set + gas-null as pre-existing behaviour this proposal doesn't change - also works. But I'd lean toward the first.

Either way, my opinion is the doc and the test need to agree.

proposals/contract-call-query-gas-estimation.md
```diff
@@ -0,0 +1,291 @@
...
+- `MirrorNodeContractEstimateQuery` is already exposed in the SDKs that have
+  shipped it (e.g., hiero-sdk-js, hiero-sdk-java). The helpers above wrap
+  that existing functionality so callers do not have to compose it
+  themselves.
+- The helper methods do not require the query to be frozen. They read the
+  query's existing `contractId`, `functionParameters`, and `sender` fields
+  exactly as a subsequent `execute(client)` would.
+- The `enabled` parameter on `setAutomaticGasEstimation` is non-nullable; the
+  flag is a tri-no-state boolean (the default `false` is taken when the
+  setter has never been called).
```
rwalworth: **nitpick:** "tri-no-state boolean" looks like a copy-paste artifact to me - I think it's just a boolean with a default of `false`. Suggested rewrite:

> The `enabled` parameter on `setAutomaticGasEstimation` is non-nullable. The flag defaults to `false` (i.e. when `setAutomaticGasEstimation(...)` has never been called).

(Becomes a moot comment if you take the suggestion to drop the flag.)

proposals/contract-call-query-gas-estimation.md
rwalworth, line 1: **issue (blocking):** I'd really like to see a "Rationale / Why not just require `setGas()`?" section added.

The pushback you flagged is going to come up from every reviewer who hasn't read [hiero-sdk-js#2848](https://github.com/hiero-ledger/hiero-sdk-js/issues/2848), in my experience. Right now the doc reads to me as a fix description but not a justification - it explains *what* changes and *how*, but not *why this rather than "require setGas() like everyone else"*. My view is that two paragraphs near the top would make this design durable.

Points I'd hit (these are my opinions on what makes the design defensible):

1. **`ContractCallQuery` is a read query, and from what I can see EVM read calls don't take caller gas.** ethers.js `staticCall`, viem `readContract`, web3.js `.call()` all execute reads without a user-supplied gas value. Gas is required by Hedera here only because of how `COST_ANSWER` is priced (`gas × tinybars-per-gas`), not because the EVM semantically needs it from the caller for a local call.
2. **The Hedera JSON-RPC relay already auto-estimates** for `eth_call` - to me that's the whole reason the issue reporter's contract works on Hashscan and fails in the native SDK. My read is the native SDKs are diverging from their own relay.
3. **Mode A (manual gas) is unchanged.** Callers who do offline gas estimation see no behavioural change and pay no overhead.

Without this section, my prediction is the design could keep getting re-litigated.

rwalworth (changes requested):
Thanks for putting this together @ddeskov-limechain - and honestly, thanks for surfacing the "do we even need this?" question explicitly, because that's the right thing to pressure-test before this turns into 7 SDK implementations.

I went and verified the failure mode in `hiero-sdk-js` (`Query.js _beforeExecute` calls `getCost()` when `_queryPayment` is null --> `CostQuery` wraps the query --> `gas: this._gas` serialises a null int64 as 0 --> consensus node rejects with `INSUFFICIENT_GAS`, and the stack trace points at `ContractCallQuery._mapStatusError`, so users see what looks like *their* call failing). From what I can see, your diagnosis is correct, and the consensus node is behaving exactly per `services_contract_call_local.proto`. So in my view **this is unambiguously an SDK bug** - that's not the question I'd debate.

The question you raised - whether the right response is to fix the bug *and* add estimation, or just fix the bug and tell everyone to call `setGas()` - is the interesting one. My opinion: **don't push back on the issue.** Here's where I'm coming from:

`ContractCallQuery` is a *read* query - local execution, no state change. To me, the EVM-ecosystem analogue is `eth_call`, and from what I've seen `eth_call` does not take gas from the caller in any major library:

- ethers.js v6 `staticCall` - no caller gas.
- viem `readContract` - no caller gas.
- web3.js `.call()` - no caller gas.
- Hedera's own JSON-RPC relay - estimates internally; I think this is exactly *why* the issue reporter's contract works on Hashscan and breaks in the native SDK.

So my read is: when an EVM developer picks up the native Hiero SDK, calls `ContractCallQuery().setContractId(...).setFunction(...)`, and it blows up with `INSUFFICIENT_GAS`, they are not encountering an ecosystem norm - they're encountering a Hedera-specific quirk caused by how `COST_ANSWER` is priced (`gas × tinybars-per-gas`), pushed onto them by an SDK that fails silently and reports the failure misleadingly. To me, telling those developers "you should always call `setGas()` first like every other chain" doesn't really hold up - most other chains don't ask for gas on reads at all. And the relay already proves we can do better.

**My bottom line on the validity question:** I think the bug fix (fail-fast with a clear error before any network call) is non-negotiable and worth shipping by itself. Estimation on top of that, in my view, is a real ergonomic improvement that closes a gap with the relay and with the EVM ecosystem. My recommendation would be: don't weaken the proposal.

That said, **my view is the design is over-built for what it needs to do**, and there are two pieces of API surface I'd personally push back on before approving. I left comments for those concerns inline, along with some other smaller things.

proposals/contract-call-query-gas-estimation.md
ddeskov-limechain: I updated the doc

PR #276, opened 2026-07-22
feat: add network node health report proposal

Reccetech:
## Summary

A new proposal introducing a \`NodeHealth\` data class and \`client.getNetworkHealth()\` method that surfaces the SDK's existing per-node health state as a structured public API. Also standardises \`ping()\` and \`pingAll()\` across all SDKs — Go and Java already have these; JS has neither.

**Primary motivation — external signing:** Developers who sign transactions outside the SDK (HSMs, offline signers, air-gapped infrastructure) must embed \`nodeAccountID\` in \`TransactionBody\` before signing. Because the signature covers the full serialised bytes, the node choice is locked in at signing time. If the selected node is unhealthy, the transaction fails on submission and cannot be re-routed without a new signing operation. \`getNetworkHealth()\` lets these workflows select a healthy node before committing to a signature.

**Key design decisions:**
- \`getNetworkHealth()\` makes no network calls — reads the SDK's existing internal health state only
- \`NodeHealth\` exposes \`nodeId\` (numeric, matching HashScan / mirror node explorer), \`accountId\`, and \`description\` (from the address book) alongside health metrics
- \`ping()\` accepts either \`nodeId\` or \`accountId\` for flexible targeting
- Notes dependency on \`AccountBalanceQuery\` as the current ping probe, with guidance for replacement if that class is removed
- Calls out one JS SDK bug to fix as a prerequisite (\`_badGrpcStatusCount\` never incremented)

**Related proposals:**
- [AccountBalanceQuery Deprecation](./proposals/account-balance-query-deprecation.md)
- [AccountBalanceQuery Mirror Node Migration](./proposals/account-balance-query-mirror-node-migration.md)

## Test plan
- [ ] Review proposal content and API design
- [ ] Verify meta-language API blocks conform to \`guides/api-guideline.md\`
- [ ] Confirm \`NodeHealth\` field mapping to SDK internals is accurate per SDK
- [ ] Verify test plan items align with TCK requirements

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/proposals/network-node-health-report.md
+++ b/proposals/network-node-health-report.md
@@ -1,0 +1,218 @@
+# Network Node Health Report
+
+## Summary
+
+The Hiero SDKs track per-node health state internally — backoff windows, readmit timestamps, bad gRPC status counts, and use counters — but expose none of it to developers. This proposal surfaces that state through a `NodeHealth` data class and `client.getNetworkHealth()`. It also standardises `ping()` and `pingAll()` across all SDKs; Go and Java already have these, but JS has neither.
+
+### The external signing problem
+
+Developers who sign transactions outside the SDK — using HSMs, offline signers, or custom signing services — must set `nodeAccountID` in `TransactionBody` before signing. Because the signature covers the full serialised `TransactionBody` bytes, the node choice is irrevocably locked in at signing time. If that node is unhealthy, the transaction fails on submission and cannot be re-routed; a new signing operation is required for each new `nodeAccountID`, which may mean a round-trip to an HSM or air-gapped environment.
+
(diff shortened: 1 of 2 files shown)


proposals/network-node-health-report.md
```diff
@@ -0,0 +1,218 @@
...
+NodeHealth {
+    @@immutable nodeId: int32
+    @@immutable accountId: AccountId
+    @@immutable @@nullable description: string
+    @@immutable isHealthy: bool
+    @@immutable currentBackoff: Duration
+    @@immutable remainingBackoff: Duration
+    @@immutable useCount: int64
+    @@immutable badGrpcStatusCount: int64
+}
```
steven-sheehy, line 39: I don't think we have to expose all the internal details. We just need a filtered list of nodes that are considered healthy by the SDK to support offline signing. So it could just have the same return data as `client.getNetwork()`.

PR #281, opened 2026-08-14
feat: add MirrorNodeTokenBalanceQuery proposal

Reccetech:
## Summary

This proposal introduces `MirrorNodeTokenBalanceQuery`: a new SDK query class for fetching token balances from the mirror node REST API, as a follow-on to `MirrorNodeAccountBalanceQuery`.

### Key design decisions

- `accountId` is required; scopes the query to a single account's finite set of token associations
- `tokenId` is optional; when set, makes a single request with no pagination (useful for single-token lookup)
- SDK auto-paginates following `links.next` (`token.id=gt:` cursor) until null when no `tokenId` filter is set
- Returns `List<MirrorNodeTokenBalance>` — each item contains `tokenId`, `balance`, and `decimals` consolidated into a single object
- Replaces the existing split-map pattern (`TokenBalanceMap` + `TokenDecimalMap`) with a unified type

### Related proposals

- [MirrorNodeAccountBalanceQuery](./proposals/account-balance-query-mirror-node-migration.md)
- [AccountBalanceQuery Deprecation](./proposals/account-balance-query-deprecation.md)

Signed-off-by: Keith Kowal <keith.kowal@hashgraph.com>

### DIFF
--- a/proposals/mirror-node-token-balance-query.md
+++ b/proposals/mirror-node-token-balance-query.md
@@ -21,21 +21,25 @@ This is a follow-on to `MirrorNodeAccountBalanceQuery` and should be implemented
 
 ### `MirrorNodeTokenBalance`
 
-A read-only data class representing a single token balance for an account.
+A read-only data class representing a single token relationship entry as returned by the mirror node. Fields map 1:1 to the mirror node `TokenRelationship` schema — no fields are filtered or projected. A developer using the SDK gets the same data as one calling the REST API directly.
 
 ```
 @@finalType
 MirrorNodeTokenBalance {
-    @@immutable tokenId: TokenId
-    @@immutable balance: Long    // raw balance in the token's smallest denomination
-    @@immutable decimals: int    // number of decimal places for the token
+    @@immutable tokenId:                TokenId
+    @@immutable balance:                Long      // raw units in the token's smallest denomination
+    @@nullable @@immutable decimals:    int       // decimal places; null for NFTs
+    @@nullable @@immutable automaticAssociation: Boolean
+    @@nullable @@immutable createdTimestamp:     String    // e.g. "1234567890.000000000"
+    @@nullable @@immutable freezeStatus:         String    // NOT_APPLICABLE | FROZEN | UNFROZEN
+    @@immutable kycStatus:             String    // NOT_APPLICABLE | GRANTED | REVOKED
 }
 ```
 
-`balance` is the raw on-chain value in the token's smallest unit. To get a human-readable amount, divide by `10^decimals`. This matches the behaviour of the existing `AccountBalance.tokens` map, which also returned raw balances.
+`balance` is the raw on-chain value. For fungible tokens, divide by `10^decimals` for a human-readable amount. For NFTs, `balance` is the count of NFTs held and `decimals` is null.
 
 **Design note — consolidation of balance and decimals:**
-The existing `AccountBalance` type split token data across two separate maps: `tokens: TokenBalanceMap` (balance keyed by `TokenId`) and `tokenDecimals: TokenDecimalMap` (decimals keyed by `TokenId`). `MirrorNodeTokenBalance` consolidates both into a single object per token — balance and decimals are always available together. SDK teams implementing this type do not need to replicate the split-map pattern.
+The existing `AccountBalance` type split token data across two separate maps: `tokens: TokenBalanceMap` (balance keyed by `TokenId`) and `tokenDecimals: TokenDecimalMap` (decimals keyed by `TokenId`). `MirrorNodeTokenBalance` consolidates both into a single object per token alongside the full token relationship context. SDK teams implementing this type do not need to replicate the split-map pattern.
 
 ---
 
@@ -116,7 +120,7 @@ GET /api/v1/accounts/{accountId}/tokens?limit=100&order=asc
 GET {nextPage}   // use the links.next URL verbatim — do not reconstruct it
 ```
 
-**Mirror response shape per token:**
+**Mirror response shape per token (`TokenRelationship` schema):**
 ```json
 {
   "token_id": "0.0.12345",
@@ -129,7 +133,19 @@ GET {nextPage}   // use the links.next URL verbatim — do not reconstruct it
 }
 ```
 
-Parse `token_id` → `TokenId`, `balance` → `Long`, `decimals` → `int`. Map `links.next` → `MirrorNodeTokenBalancePage.next` (null when absent). The remaining fields (`automatic_association`, `created_timestamp`, `freeze_status`, `kyc_status`) are out of scope for a balance query and are not exposed on `MirrorNodeTokenBalance`.
+All fields are mapped to `MirrorNodeTokenBalance` with no filtering:
+
+| Mirror node field       | SDK field             | Type                        |
+|-------------------------|-----------------------|-----------------------------|
+| `token_id`              | `tokenId`             | `TokenId`                   |
+| `balance`               | `balance`             | `Long`                      |
+| `decimals`              | `decimals`            | `int` (nullable for NFTs)   |
+| `automatic_association` | `automaticAssociation`| `Boolean` (nullable)        |
+| `created_timestamp`     | `createdTimestamp`    | `String` (nullable)         |
+| `freeze_status`         | `freezeStatus`        | `String` (nullable)         |
+| `kyc_status`            | `kycStatus`           | `String`                    |
+
+Map `links.next` → `MirrorNodeTokenBalancePage.next` (null when absent). This type corresponds directly to the mirror node's `TokenRelationship` OpenAPI schema — SDK teams may choose to generate it from the spec rather than hand-writing it.
 
 ### Free query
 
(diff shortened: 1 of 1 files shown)


proposals/mirror-node-token-balance-query.md
```diff
@@ -0,0 +1,218 @@
...
+Use `limit=100` (not `limit=1`) so that `links.next` is null when the single result is returned. With `limit=1`, the mirror node echoes a non-null `links.next` even for a single-item result (see pagination note below). In either case, the SDK does not follow `links.next` when `tokenId` is set.
+
+**Endpoint (all tokens, first page):**
+```
+GET /api/v1/accounts/{accountId}/tokens?limit=100&order=asc
+```
+
+**Pagination loop (no `tokenId` filter only):**
+
+The SDK follows `links.next` until `links.next` is null. The SDK requests 100 items per page. The cursor is embedded in `links.next` by the mirror node as a `token.id=gt:{last_id}` cursor — the SDK uses the URL verbatim rather than constructing it.
```
steven-sheehy: The SDK should not paginate for the user internally and return all tokens. There can be an unbounded number of tokens associated to an account. Doing automatic queries for all tokens when the user doesn't even necessarily want it would place a burden on the mirror node and potentially on the memory of the client. It would be slow and resource intensive. Pagination should be left to the clients to do if the initial 100 do not have what they need.

proposals/mirror-node-token-balance-query.md
```diff
@@ -0,0 +1,218 @@
...
+```
+GET /api/v1/accounts/{id}/tokens?limit=100&order=asc
+→ { tokens: [...100 items...], links: { next: "/api/v1/accounts/{id}/tokens?limit=100&token.id=gt:0.0.500" } }
+GET /api/v1/accounts/{id}/tokens?limit=100&token.id=gt:0.0.500&order=asc
+→ { tokens: [...remaining...], links: { next: null } }
+```
+
+**Important — `links.next` behavior when `tokenId` filter is set:**
+
+When `setTokenId` is used, the SDK must NOT follow `links.next`. The mirror node echoes back a non-null `links.next` with the same `token.id=` filter even when the result fits within the page limit (e.g., `limit=1` returning one result). That echoed link is a filter reference, not a pagination cursor — following it loops indefinitely returning the same token. When `setTokenId` is set, `execute` is always a single request with no pagination regardless of `links.next`.
```
steven-sheehy: Not sure what you're implying here. The unique key of a token relationship is account plus token. If they provide a request like `/api/v1/accounts/0.0.1000/tokens?token.id=0.0.1001` it will always return one or zero results and links.next=null. Here's an example: https://mainnet.mirrornode.hedera.com/api/v1/accounts/1196259/tokens?token.id=0.0.127877

proposals/mirror-node-token-balance-query.md
```diff
@@ -0,0 +1,218 @@
...
+### Eventual consistency
+
+The mirror node reflects consensus state with a small lag (typically seconds). Token balances read immediately after a transfer may still show pre-transfer values. This should be documented in SDK release notes.
+
+### Response Codes
+
+No consensus node response codes apply. Mirror node HTTP errors:
+
+- `400 Bad Request` — invalid ID format. Surface as an SDK-appropriate error. Do not retry.
+- `404 Not Found` — account does not exist. Surface as an SDK-appropriate error. Do not retry.
```
steven-sheehy, line 167: It should not return 404 for account not found, just an empty list.

proposals/mirror-node-token-balance-query.md
```diff
@@ -0,0 +1,218 @@
...
+
+A read-only data class representing a single token balance for an account.
+
+```
+@@finalType
+MirrorNodeTokenBalance {
+    @@immutable tokenId: TokenId
+    @@immutable balance: Long    // raw balance in the token's smallest denomination
+    @@immutable decimals: int    // number of decimal places for the token
+}
```
steven-sheehy, line 36: I'm not sure why we keep rewriting MN APIs with new methods and responses in the SDK. We have an OpenAPI specification that can generate response models automatically. We should investigate ways to reuse that instead of creating duplicate SDK APIs.

steven-sheehy (changes requested):

proposals/mirror-node-token-balance-query.md
hendrikebbers: Oh, is it not an RCP call but a REST call? So far we do not have any REST calls in any SDK. Based on that this would be something totally new. The API (Including call with Client instance does not make much sense than since you do not need to sign anything (with operator). In that case I would favor to discuss how an integration and architecture of something like this might look like

proposals/mirror-node-token-balance-query.md
Reccetech: Agreed. Redesigned execute() to return a single page (MirrorNodeTokenBalancePage) with an exposed next cursor rather than auto-consuming all pages internally. The caller controls pagination.

Changes:
- New MirrorNodeTokenBalancePage type with tokens (up to 100) and next cursor (null on last page)
- execute() makes exactly one HTTP request
- New setNextPage(String) setter to fetch subsequent pages using the links.next URL verbatim.
- accountId not required when setNextPage is set (account is embedded in the cursor URL)
- setNextPage and setTokenId together throw before any network call
- SDK examples updated to show both single-page and caller-driven all-pages patterns

proposals/mirror-node-token-balance-query.md
Reccetech: With the redesign from Comment 1 (caller-controlled pagination), the SDK no longer follows links.next automatically in any case. For the single-token filter we use limit=100, which ensures links.next is null. This concern is then probably resolved by these changes.
Agree?

proposals/mirror-node-token-balance-query.md
Reccetech: When I tested /api/v1/accounts/{id}/tokens the endpoint returns 404 for a non-existent account, not an empty list:

GET /api/v1/accounts/0.0.999999999         → 404 Not found
GET /api/v1/accounts/0.0.999999999/tokens  → 404 Not found

Thoughts?

proposals/mirror-node-token-balance-query.md
Reccetech: The proposal has been updated to map 1:1 to `TokenRelationship` . A developer using the SDK gets exactly the same data as one calling the REST API directly.

Does this address the concern?

hendrikebbers: My concerns still exist. See my comment at https://github.com/hiero-ledger/hiero-sdk-java/pull/2883

Reccetech: @hendrikebbers 
Looking at your PR #2883 review, the concerns you raise (client lifecycle, executor integration, transport configurability, Android compatibility) apply to the existing HTTP pattern already shipped in FeeEstimateQuery, and RegisteredNodeAddressBookQuery — not to anything specific to this proposal. This proposal follows that same, already-established pattern.

A fix as you propose would mean refactoring the SDK's Mirror Node HTTP handling across every existing query class, which is a larger architectural effort than this proposal's scope. I'd suggest tracking that as its own cross-cutting initiative.

ddeskov-limechain:
FYI on the JS side: to make the release 77 cutoff, hiero-ledger/hiero-sdk-js#4306 ships a single-token form of `MirrorNodeTokenBalanceQuery` ahead of this proposal. It is marked in the JSDoc and CHANGELOG as subject to change once this proposal is finalized, and hiero-ledger/hiero-sdk-js#4355 tracks aligning it.

Delta against the current draft here, so nobody is surprised later:

| | JS now | This proposal |
|---|---|---|
| `setTokenId` | required | optional filter |
| `setNextPage` / cursor | none | `next` cursor, caller-driven paging |
| Return type | one `MirrorNodeTokenBalance` (`tokenId`, `balance`, `decimals`) | `MirrorNodeTokenBalancePage` with the full `TokenRelationship` fields |
| Account holds no relationship | zero balance, decimals 0 | empty page |
| Unknown account | not distinguished | error on 404 |

The input side is a strict subset of the proposal, so migrating is mostly the return type. Java and Go are holding off on the token query until this merges, so they implement it once.

ddeskov-limechain:
A note from the TCK side (hiero-ledger/hiero-sdk-tck#715).

This proposal conflicts with #286, which the SDK teams now treat as the source of truth for mirror REST:

- **Cursor:** #286 says a paginated result must never hand the caller a raw cursor (the `links.next` confused-deputy path). This proposal exposes `next` and `setNextPage(String)`.
- **Retry:** #286 pins the retryable statuses to 408, 429, 500, 502, 503 and 504. This proposal retries 500 and 503 only.
- **API shape:** #286 lists `MirrorNodeTokenBalanceQuery` as an existing query with no public signature change, meaning the single-token version (required `setTokenId`, one call).

For the AccountBalanceQuery Stage 2 release, the TCK now tests that single-token query (hiero-ledger/hiero-sdk-tck#720). The open token points are tracked in hiero-ledger/hiero-sdk-tck#716.

One more ask: #282 fixes the unknown-account text in the merged migration proposal (`INVALID_ACCOUNT_ID`, which every SDK and the TCK already use), but it is stacked on this PR. Could #282 be rebased onto `main` so it can merge on its own?

PR #282, opened 2026-08-19
fix: update MirrorNodeAccountBalanceQuery proposal

Reccetech:
## Summary

- Rename document title to `MirrorNodeAccountBalanceQuery`
- Fix non-existent account behaviour: `execute()` now throws `INVALID_ACCOUNT_ID` when the mirror node returns an empty `balances` array, instead of silently returning `hbars = 0`. A genuine 0-HBAR account returns `[{ balance: 0 }]` and is handled without error — the two cases are unambiguously distinguishable from the same response
- Add both mirror node response shapes (account exists vs. does not exist)
- Remove the "Why token balances are not returned" subsection (was inaccurate — the `/api/v1/balances` endpoint does return token data; explanation moved to new section)
- Add "Token Balances" section at the end, linking to the follow-on `MirrorNodeTokenBalanceQuery` proposal
- Add test 6: genuine 0-HBAR account returns `hbars = 0` without error
- Fix EVM address example: 39 → 40 hex characters
- Fix `const` redeclaration in EVM/alias example (would be a JS syntax error)
- Rename "Transaction Retry" heading to "Retry Policy"
- Update stale cross-references throughout

## Related

- Follow-on proposal: #281 (MirrorNodeTokenBalanceQuery)

### DIFF
--- a/proposals/account-balance-query-mirror-node-migration.md
+++ b/proposals/account-balance-query-mirror-node-migration.md
@@ -1,4 +1,4 @@
-# AccountBalanceQuery Mirror Node Migration
+# MirrorNodeAccountBalanceQuery
 
 ## Summary
 
@@ -11,6 +11,7 @@ The new class follows the naming and structural convention already established i
 **Related references:**
 - [Hedera blog: Migrating from AccountBalanceQuery](https://hedera.com/blog/migrating-from-accountbalancequery-what-you-need-to-know/)
 - [Mirror node REST docs: GET /api/v1/balances](https://docs.hedera.com/hedera/sdks-and-apis/rest-api#balances)
+- [Follow-on proposal: MirrorNodeTokenBalanceQuery](./mirror-node-token-balance-query.md)
 
 ---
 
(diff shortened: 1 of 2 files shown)


Dosik13:
It looks good but there are some things I want to address:

1 - I don't think this should be tied to `MirrorNodeTokenBalanceQuery` yet. Right now the doc is self-contained about token balances - line 21 points inward, line 101 carries the reason. This PR swaps that for two forward references to #281: the new bullet on line 14, and a `## Token Balances` section that's a pointer plus a summary of #281's design.
That feels like the wrong direction of dependency. This proposal is merged with two implementations already built on it; `MirrorNodeTokenBalanceQuery` exists in no SDK yet, and #281 has been redesigned twice in the last week. 

2 - An empty `balances` array doesn't unambiguously mean the account doesn't exist. It means the mirror node has no row for that ID *yet* — which also covers an account it hasn't ingested. The Eventual consistency section two subsections down says that lag exists, so the doc argues both sides.
The concrete failure is create-then-read: an account created a moment ago returns `INVALID_ACCOUNT_ID` even though it exists. And "Do not retry" on line 145 makes that unrecoverable, when retrying is the only remedy.

Three wording changes fix it:

- say the mirror node has no record *yet*, rather than asserting the account is invalid;
- say a caller querying an account it just created should retry, not treat the first failure as final;
- say the status is the SDK's mapping of an empty result, not something the mirror node returned.

PR #283, opened 2026-08-21
Add Mirror Node HTTP client architecture proposal

Reccetech:
## Summary

Proposes a single `Client`-owned HTTP transport for **all** mirror-node REST call sites across the Hiero SDKs — queries, entity-id resolution, address book, fee estimate, contract-call / estimate-gas, and populate helpers — replacing today's per-class static clients, one-off `fetch` / `requests` / `URLSession` calls, and new-client-per-call patterns.

It covers two things:
- **Transport ownership:** a language-native transport SPI, `Client.close()` closing HTTP, Android-safe defaults (no `java.net.http` on the Android class-load path), identity/headers (`User-Agent` + caller headers), one retry policy, and per-attempt timeout semantics.
- **Behavioral alignment:** converging the query behavior that currently drifts per language (auto-freeze, chunk aggregation, missing-entity handling, lossless int64/uint64 parsing, timeouts, wire format, error envelope).

Directly addresses the architecture concern raised in [hiero-sdk-java#2883](https://github.com/hiero-ledger/hiero-sdk-java/pull/2883) and the follow-on to [#281](https://github.com/hiero-ledger/sdk-collaboration-hub/pull/281).

## Details

- New APIs: `MirrorNodeHttpHeaders` and a `MirrorNodeHttpTransport` SPI (with an explicit transport contract: non-2xx returned not thrown, concurrency-safe, fail-fast after close), plus `Client` header/transport setters.
- REST call-site/endpoint table, retry classification + jittered backoff (honoring `Retry-After`), and a per-attempt timeout model.
- A 28-item test plan mapped to the TCK and a non-blocking 6-phase rollout.

## Decision asked of reviewers

1. Finalize updated architecture
2. Do we move forward with MirrorNodeAccountBalanceQuery and MirrorNodeToken Query with the existing architecture and then refactor, or implement with the new architecture.

Signed-off-by: Keith Kowal <keith.kowal@hashgraph.com>

### DIFF
--- a/proposals/mirror-node-http-client-architecture.md
+++ b/proposals/mirror-node-http-client-architecture.md
@@ -1,0 +1,394 @@
+# Updating the Mirror Node HTTP Client Architecture
+
+## Summary
+
+Mirror-node REST is now a first-class SDK transport. `FeeEstimateQuery`, `RegisteredNodeAddressBookQuery`, `MirrorNodeAccountBalanceQuery`, contract-call / estimate-gas, entity-id resolution, and the in-flight `MirrorNodeTokenBalanceQuery` each open HTTP themselves — a static `HttpClient`, a one-off `fetch` / `requests.post` / `URLSession`, or a new client per call.
+
+The failure modes are documented with [hiero-sdk-java#2883](https://github.com/hiero-ledger/hiero-sdk-java/pull/2883). The same gaps appear, in different shapes, in every Hiero SDK. The SDKs also have differences on **query behavior** (freeze, chunks, missing-entity, integer precision, timeouts). This proposal covers both: one owned transport, and one product contract.
+
+**Date Submitted:** 2026-08-21
+
+**Related references:**
+- [hiero-sdk-java#2883](https://github.com/hiero-ledger/hiero-sdk-java/pull/2883) — Hendrik's review
+- [sdk-collaboration-hub#281](https://github.com/hiero-ledger/sdk-collaboration-hub/pull/281)
+- [Mirror Node ingress / local defaults](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/mirror-node-ingress-endpoint-standardization.md) — URL rewriting; out of scope here
+- [MirrorNodeAccountBalanceQuery](./account-balance-query-mirror-node-migration.md)
+- [MirrorNodeTokenBalanceQuery](./mirror-node-token-balance-query.md)
+- [Mirror-node eventual consistency](./mirror-node-eventual-consistency.md)
+- [hiero-sdk-js#4336](https://github.com/hiero-ledger/hiero-sdk-js/issues/4336) — JS JSON int64 precision
+- HIP-1261 — fee-estimate retry / chunk aggregation baseline
+
+---
+
+## Problems
+
+### 1. Duplicated, unowned clients
+
+Java:
+
+```java
+private static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();
+```
+
+appears in `FeeEstimateQuery`, `RegisteredNodeAddressBookQuery`, and the in-flight balance-query classes. `EntityIdHelper.performQueryToMirrorNodeAsync` builds a **new** `HttpClient` per call. None participate in `Client.close()`. (`HttpClient` is `AutoCloseable` since Java 21.)
+
+Each Java static client has its own selector thread and pool. "Leaks transport threads / blocks JVM shutdown" is overstated: on JDK 11+ the selector thread is a daemon. The real bug is ownership — a short-lived `Client` should not leave process-lifetime pools, and `close()` should be enough.
+
+### 2. Executor / thread-context
+
+`RegisteredNodeAddressBookQuery.executeAsync` uses `CompletableFuture.supplyAsync(..., client.executor)` while I/O still runs on `HttpClient`'s threads. `FeeEstimateQuery.executeAsync` uses `sendAsync().whenComplete(...)` and retries with `Thread.sleep` on that callback thread — no `client.executor` at all. Tracing / MDC set on `Client` do not apply. No production bug is tied to this yet.
+
+### 3. Transport configurability
+
+Bare `HttpClient.newHttpClient()` / `fetch()` / `http.Client` cannot take a corporate proxy, custom truststore or mTLS, an `Authorization` header, pool / HTTP-version settings, an SDK `User-Agent`, or tracing interceptors. Real for private-mirror / enterprise users; no reported production blocker found. These should be *possible* once a transport exists; they do not all need a cross-SDK setter in v1.
+
+### 4. Android (Java, confirmed)
+
+`java.net.http` is not on Android libcore at any API level. Classes with a static `HttpClient` throw `NoClassDefFoundError` on load. Android is a maintained target (README badge, `example-android/`, CI, changelog). The SDK already tells Android users to swap `grpc-okhttp` for Netty. HTTP needs the same shape of fix. This is broken today.
+
+`EntityIdHelper`'s javadoc says "API 33+". That is wrong: the class fails on every Android API, not only below 33.
+
+### 5. Query behavior drift
+
+Every SDK exposes the same query classes, but each one implements them differently — so the *same* call returns *different* results depending on the language. Unifying the HTTP transport (Problems 1–4) does not fix this; the normative behavior each SDK should converge on is in [Behavioral alignment](#behavioral-alignment-beyond-transport). The divergences today:
+
+- **Java** does not auto-freeze on execute and has no chunk loop (`setTransaction(Transaction)` snapshots via `makeRequest()`; unfrozen txs fail there).
+- **Rust** estimates the first chunk only and sends `highVolumeThrottle` / `application/x-protobuf`.
+- **JS** maps `/balances` `[]` to 0 HBAR and parses money fields with `response.json()` (IEEE-754; values above `2^53−1` truncate). `AccountId.populateAccountNum` / `populateAccountEvmAddress` sleep 3 s then `fetch` with no retry or timeout. `ContractId.populateAccountNum` is the same `fetch` **without** the sleep.
+- **Swift** parses fees via `JSONSerialization` → `NSNumber.uint64Value` (same precision *risk* as JS; not separately reproduced) and derives chunk `total` by summing per-chunk totals rather than from the aggregated formula.
+- **Go** fee estimate passes timeout `0` into the retry helper (no per-request timeout).
+- **Python** always sends `high_volume_throttle=0` and mutates the caller's transaction during the chunk loop.
+
+---
+
+## Goals
+
+1. **One HTTP layer per `Client`.** All mirror REST (queries, entity-id, address book, fee estimate, contract-call, populate) goes through it.
+2. **`Client.close()` closes HTTP.** In-flight calls are waited out, then aborted (same budget as gRPC shutdown). After close, new calls fail; they do not construct a new client.
+3. **Default transport is platform-safe.** Java must not load `java.net.http` on the Android class-init path.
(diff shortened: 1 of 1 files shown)


proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+    MirrorNodeHttpResponse execute(request: MirrorNodeHttpRequest)
+
+    // Client.close(). After close, execute() fails fast (see error contract).
+    // In-flight: wait up to closeTimeout, then abort. Idempotent (second close is a no-op).
+    close(closeTimeout: duration): void
+}
+
+@@finalType
+MirrorNodeHttpRequest {
+    @@immutable method: string          // "GET" | "POST"
```
hendrikebbers: Should be an enum

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+
+@@finalType
+MirrorNodeHttpResponse {
+    @@immutable statusCode: int
+    @@immutable body: bytes
+    @@immutable headers: map<string, string>
+}
+```
+
+Swift may keep `URLSession`; JS a custom `fetch`. They do not need a class literally named `MirrorNodeHttpTransport`.
```
hendrikebbers: The benefit of the described approach is that all SDKs work similar and the underlying native transport can be changed. Let's assume Swift deprecated the current HTTP layer since they have a new API. with our abstraction a user will not be affected. That's why I really like your suggested abstraction and would like to use it as default in all SDKs

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+- **After `close()`**, `execute()` fails fast with the language-native "client closed" error and MUST NOT construct a new client or open a new connection.
+- **Thread-safety of `close()`.** `close()` is idempotent and safe to race with in-flight `execute()` calls; those calls either complete within `closeTimeout` or are aborted with a transport-error.
+
+Query classes keep their existing public error types. A transport-error or a retry-exhausted non-2xx is wrapped into that per-query type at the helper boundary, preferring the mirror `_status` envelope for the message (see [Response Codes](#response-codes)).
+
+| SDK | Default transport |
+|---|---|
+| Java | Android-safe (OkHttp *or* `HttpURLConnection` + SPI). **Not** `java.net.http` on the Android class-load path. See Open Questions. |
+| JS | Platform `fetch` |
+| Go | Reused `http.Client` on `Client` |
```
hendrikebbers: I would never but it in Client since this class is already way tu big

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+}
+```
+
+Swift may keep `URLSession`; JS a custom `fetch`. They do not need a class literally named `MirrorNodeHttpTransport`.
+
+#### Transport contract
+
+The SPI is the transport boundary; **classification and retry live in the shared helper above it, never in the transport.**
+
+- **Non-2xx is not an error at the SPI.** `execute()` returns any completed HTTP exchange — including `4xx` / `5xx` — as a `MirrorNodeHttpResponse` with its `statusCode`. The transport MUST NOT throw on status code alone. The helper reads `statusCode` and decides retry vs. surface (see [retry table](#internal-changes)).
```
hendrikebbers: "The helper reads ..." - What dos you mean by helper? I agree that execute should always return a HTTP response if it receives one from the server. Only if the connection can not be established we need to discuss how that is handled.

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+
+---
+
+## New APIs
+
+Types below use a language-neutral notation: `@@finalType` = immutable value type, `abstraction` = interface/trait/protocol to implement, `@@immutable` = read-only field, `@@nullable` = may be absent, `@@async` = returns a future/promise/task. Concrete names and shapes are language-native (see the default-transport table).
+
+### `MirrorNodeHttpHeaders`
+
+Extra headers the `Client` sends on every mirror HTTP request. The SDK `User-Agent` is always present. A caller `User-Agent` is appended (space-separated), not replaced.
```
hendrikebbers: > A caller `User-Agent` is appended (space-separated), not replaced.

Why?

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+5. **One retry policy** for transient HTTP.
+6. **Existing knobs apply.** `requestTimeout` (and `client.executor` where the SDK has one) cover HTTP as they cover gRPC.
+7. **Consistent query behavior across SDKs.** Freeze, chunks, missing-entity, lossless integers, timeouts, and wire format behave the same way in every language.
+
+## Non-goals
+
+- URL / port / localhost rewriting — [ingress proposal](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/mirror-node-ingress-endpoint-standardization.md). This doc only records current drift.
+- A gRPC `Query` subclass for mirror REST. Queries stay plain classes.
+- One HTTP implementation or a required extra dependency in every SDK.
+- Identical proxy / mTLS / HTTP2 / interceptor setters across languages. Those stay native on the injected transport.
```
hendrikebbers: If that is the case the API needs a way to expose the native client. But I believe that can happen afterwards and should not be in scope of this definition.

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+| `ContractId` populate | GET | `/api/v1/contracts/{id}` | — | No bare `fetch` |
+| `EntityIdHelper` resolve | GET | `/api/v1/accounts/{id}` (or `/contracts/{id}`) | — | No new `HttpClient` per call |
+
+The fee-estimate path itself is defined by HIP-1261; this proposal does not fix it, only that the request goes through the shared transport with the protobuf `Content-Type`.
+
+---
+
+## Internal Changes
+
+- Default transport is created with the `Client` or lazily on first use. Closed exactly once from `Client.close()`; a second close is a no-op.
```
hendrikebbers: In general I dislike this. Maybe it is a too big change for V2 to fix this. But Client today has mostly no information that is needed for a mirror node call. Like Account or keys are not needed at all.

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+
+  **Retry safety.** Every mirror REST call in scope is read-only — `GET`s plus the fee-estimate / contract-call / estimate-gas `POST`s, none of which mutate mirror state — so retrying a `POST` on a `5xx`/timeout is safe. The transport MUST NOT assume `POST` is non-idempotent and suppress retries.
+
+  **Timeouts.** `requestTimeout` is a **per-attempt** deadline (it is `MirrorNodeHttpRequest.timeout`), not a budget across the whole retry loop. Each attempt gets the full `requestTimeout`; total wall-clock is bounded by `maxAttempts × (requestTimeout + backoff)`. A per-attempt timeout is classified as a transport failure and retried like any other. SDKs that also expose an overall deadline (e.g. Go `context`) still pass it through, and it wins when it is the tighter bound.
+
+- Query classes do not construct or hold an HTTP client, do not implement retry once the helper exists (chunk aggregation stays in the query), do not set `User-Agent`, do not sleep for mirror lag, and do not parse money fields as IEEE-754 doubles.
+- **Java:** no `java.net.http` on any class Android loads. `example-android/` must compile-and-load at least one mirror REST query.
```
hendrikebbers: Here we should check if we directly provide 2 SPI implementations for Java. Outside Android `java.net.http` is the way to go.

hendrikebbers (commented):
See my comments.

Next to that I tried to create a "as clean and minimal as possible" API for V3: https://github.com/hiero-hackers/hiero-sdk-v3-prototyping/pull/9

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+---
+
+## Goals
+
+1. **One HTTP layer per `Client`.** All mirror REST (queries, entity-id, address book, fee estimate, contract-call, populate) goes through it.
+2. **`Client.close()` closes HTTP.** In-flight calls are waited out, then aborted (same budget as gRPC shutdown). After close, new calls fail; they do not construct a new client.
+3. **Default transport is platform-safe.** Java must not load `java.net.http` on the Android class-init path.
+4. **Identity and headers.** Every request sends `User-Agent: hiero-sdk-<language>/<semver>`. Callers can add headers on `Client`.
+5. **One retry policy** for transient HTTP.
+6. **Existing knobs apply.** `requestTimeout` (and `client.executor` where the SDK has one) cover HTTP as they cover gRPC.
```
Dosik13: I think here we should make them separate - there could be one setting for the http and other for the grpc, wdyt?

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+- **Default transport policy (all SDKs, overridable via `setMirrorHttpTransport`):** verify TLS certificates against the platform trust store; follow up to 5 HTTP redirects, but drop the `Authorization` and other caller headers on a cross-origin redirect; allow plaintext `http://` when the mirror base URL is explicitly `http://` (required for the local-mirror path and the C++ hardcoded-`https://` bug), i.e. do not force-upgrade the scheme. Connection reuse / keep-alive is enabled; pool sizing stays at the platform default unless an injected transport overrides it.
+- Where the SDK has `client.executor`, **completion and retry scheduling** run on it. I/O threads may be the transport's own (same as Netty / okhttp for gRPC). JS / Python keep the platform default.
+- `User-Agent: hiero-sdk-<language>/<semver>` on every request (e.g. `hiero-sdk-java/2.51.0`). Caller `User-Agent` is appended.
+- Retry (all mirror REST, overridable via existing `maxAttempts` / backoff setters):
+
+  | Condition | Retry? |
+  |---|---|
+  | 408, 429, 500, 502, 503, 504 | Yes |
+  | Timeout, connection reset, DNS failure | Yes |
+  | 400, 404, other 4xx | No |
```
Dosik13: what about the other 5xx?

proposals/mirror-node-http-client-architecture.md
```diff
@@ -0,0 +1,394 @@
...
+Phase 1 is the only shipping break on a maintained target. Phases 2–3 are Hendrik's architecture. Phase 4 is product alignment. Local URL rewriting stays on the ingress timeline.
+
+---
+
+## Open Questions
+
+1. **Java default stack.** OkHttp (matches `grpc-okhttp`, interceptors/proxy/TLS, new JVM dependency) vs `HttpURLConnection` (no new dep) vs optional `java.net.http` in a JVM-only module. Maintainers pick in review.
+2. **`setMirrorHttpTransport` public in every language in Phase 2?** Recommendation: yes wherever a mock is useful; JS takes a `fetch` function.
+3. **User-Agent version.** Same channel each SDK already uses for its published version.
+4. **`maxAttempts` defaults.** Recommendation: standardize classification + backoff only; leave attempt counts.
```
Dosik13: I think now the maxAttempts that are used for http differ in every sdk so I think we should pick a number so that the time is the same in every sdk

proposals/mirror-node-http-client-architecture.md
Dosik13: I agree that the class is too big but like the Client the base thing the user need to work with the SDK. Like the http.Client instance could live in this class and when the Client is being closed - thats when we close the http.Client. I will make a better research and will give you some options so we can choose the best one.

hendrikebbers:
Here is how the V3 spec (draft/wip) defines it: https://github.com/hiero-hackers/hiero-sdk-v3-prototyping/blob/main/spec/base/http.md

